PR #8, opened 2015-01-08
Ignore empty lines and exceptions when parsing CSV

falaki:
* Adds dependency on slf4j
* Ignores empty records form CSVParser
* Ignores records that CSVParser throws exception
* Bumps the version to 0.1.1

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -28,6 +28,7 @@ import org.apache.spark.sql.catalyst.types.{StructType, StructField, StringType}
 import org.apache.spark.sql.sources.TableScan
 
 import scala.collection.JavaConversions._
+import scala.util.control.NonFatal
 
 case class CsvRelation protected[spark] (
     location: String,
@@ -36,7 +37,7 @@ case class CsvRelation protected[spark] (
     quote: Char,
     userSchema: StructType = null)(@transient val sqlContext: SQLContext) extends TableScan {
 
-  private val logger = LoggerFactory.getLogger(classOf[CsvRelation])
+  private val logger = LoggerFactory.getLogger(CsvRelation.getClass)
 
   val schema = inferSchema()
 
@@ -136,7 +137,7 @@ case class CsvRelation protected[spark] (
           Some(projection(row))
         }
       } catch {
-        case e: Exception =>
+        case NonFatal(e) =>
           logger.error(s"Exception while parsing line: $line. ", e)
           None
       }
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -128,8 +128,10 @@ case class CsvRelation protected[spark] (
           None
         } else {
           val tokens = records.head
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -35,6 +36,8 @@ case class CsvRelation protected[spark] (
     quote: Char,
     userSchema: StructType = null)(@transient val sqlContext: SQLContext) extends TableScan {
 
+  private val logger = LoggerFactory.getLogger(classOf[CsvRelation])
```
rxin: "classOf[CsvRelation]" -> "getClass"

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -120,12 +120,24 @@ case class CsvRelation protected[spark] (
...
-      for (index <- 0 until schemaFields.length) {
-        row(index) = tokens.get(index)
+    iter.flatMap { line =>
+      try {
+        val records = CSVParser.parse(line, csvFormat).getRecords
+        if (records.isEmpty) {
+          logger.warn(s"Ignoring empty line: $line")
+          None
+        } else {
+          val tokens = records.head
```
aarondav: Right now we expect that each element of the RDD is a line, but if we just iterated over the records and returned them, would that make everyone happy? Could also avoid the initial if check.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -120,12 +120,26 @@ case class CsvRelation protected[spark] (
...
+          var index = 0
+          while (index < schemaFields.length) {
+            row(index) = tokens.get(index)
+            index = index + 1
+          }
+          Some(projection(row))
+        }
+      } catch {
+        case e: Exception =>
```
rxin: you probably want to match NonFatal http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.util.control.NonFatal$

rxin:
this lgtm - in the next iteration if you want to improve performance, you probably want to remove the flatMap with option and just define an iterator returned by mapPartitions

aarondav:
Iterator#flatMap is implemented efficiently, is it not?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -120,12 +121,26 @@ case class CsvRelation protected[spark] (
       schemaFields: Seq[StructField],
       projection: MutableProjection,
       row: GenericMutableRow): Iterator[Row] = {
-    iter.map { line =>
-      val tokens = CSVParser.parse(line, csvFormat).getRecords.head
-      for (index <- 0 until schemaFields.length) {
-        row(index) = tokens.get(index)
+    iter.flatMap { line =>
+      try {
+        val records = CSVParser.parse(line, csvFormat).getRecords
```
aarondav: to be specific, I was referring to this:

```scala
CSVParser.parse(line, csvFormat).getRecords.map { tokens =>
  var index = 0
    row(index) = tokens.get(index)
    index = index + 1
  }
  (projection(row))
}
```

As this is already merged, though, we can do this some time later.

PR #9, opened 2015-01-14
remove redundant assembly plugin

searler:
assembly plugin also appears as freestanding plugin file


falaki:
Thanks @searler. LGTM.

PR #12, opened 2015-01-21
Adding support for DDL with schema

falaki:
We may publish this snapshot version until Spark 1.3 is released.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
@@ -17,7 +17,7 @@ package com.databricks.spark.csv
 
 import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.sources.{RelationProvider, SchemaRelationProvider}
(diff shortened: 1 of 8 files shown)


src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -56,5 +57,44 @@ class DefaultSource extends RelationProvider {
 
     CsvRelation(path, headerFlag, delimiterChar, quoteChar)(sqlContext)
   }
+
+  /**
+   * Creates a new relation for data store in CSV given parameters and user supported schema.
+   * Parameters have to include 'path' and optionally 'delimiter', 'quote', and 'header'
+   */
+  def createRelation(
```
yhuai: Seems we can consolidate the common logics in this one the above `createRelation`?

yhuai:
Seems we can also remove `JavaCsvParser`?

yhuai:
Will this PR update the README as well?

PR #14, opened 2015-01-28
wrong groupId

cvengros:
Your groupId is in fact `com.databricks`, not `com.databricks.spark` see: http://mvnrepository.com/artifact/com.databricks/spark-csv_2.10/0.1
Took me a while to figure out, with the proposed groupId it works.


falaki:
Thanks! LGTM.

PR #15, opened 2015-01-28
CSV parser wasn't recognizing custom quote and delimiters

rbolkey:
Ran into troubles parsing content with an unusual delimiter, and noticed that the delimiter wasn't being set when parsing the first line. This fixed my problem.

I'm not a scala guy, so the fix could easily be in the wrong style.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -52,6 +52,7 @@ case class CsvRelation protected[spark] (
 
     val csvFormat = CSVFormat.DEFAULT
       .withDelimiter(delimiter)
+      .withQuote(quote)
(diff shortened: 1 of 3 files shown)


JoshRosen:
I don't really know anything about `spark-csv`, but as a general best-practice it might be nice to add a regression test to [CSVSuite](https://github.com/databricks/spark-csv/blob/master/src/test/scala/com/databricks/spark/csv/CsvSuite.scala) to make sure this problem isn't re-introduced.

falaki:
@rbolkey Thanks for sending this patch. Would you please add one or two unit tests for it?

rbolkey:
Will do. Hopefully will have time tomorrow.

rbolkey:
Added a test case and data file.

falaki:
Thanks. LGTM.

PR #16, opened 2015-02-04
Allowing for empty files when the schema is defined

rbolkey:
We're aggregating multiple files with a known schema. Sometimes those files are empty. The `inferSchema` call was trying to read the first line, which seemed unnecessary when the schema was already defined. Added test case for an empty csv.


falaki:
Thanks. Would you please add another test for SQL?

toddleo:
Hi, I encoutered `java.lang.UnsupportedOperationException: empty collection` without providing schema, when reading empty files. Any suggestions to solve this?

PR #21, opened 2015-02-18
Use C for column names

vlyubin:
When visualizing the results C$i might be a better abbreviation.


falaki:
Would you add a small test for this?

vlyubin:
Ok, added a test.

falaki:
Thanks!

PR #23, opened 2015-02-19
Support for new features in Spark 1.3

falaki:
* types are no longer under catalyst
* SchemaRDD is replaced by DataFrame
* Support for writing CSV files


falaki:
@yhuai You have already reviewed most of this. Would you please take a final look?

PR #25, opened 2015-03-07
Make the package work against spark 1.3, add some new stuff

vlyubin:
Now it works against spark 1.3 RC2, and adds:
* INSERT OVERWRITE
* DefaultSource also extends CreatableRelationProvider

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,14 +60,13 @@ OPTIONS (path "cars.csv", header "true")
 ```
 
 ### Java API
-CSV files can be read using functions in CsvUtils.
+CSV files can be read using functions in JavaCsvParser.
 
 ```java
-import com.databricks.spark.csv.CSVUtils;
+import com.databricks.spark.csv.JavaCsvParser;
 
-JavaSchemaRDD cars = (new CsvUtils()).setUseHeader(true).csvFile(sqlContext, "cars.csv");
+DataFrame cars = (new JavaCsvParser()).withUseHeader(true).csvFile(sqlContext, "cars.csv");
 ```
 
 ## Building From Source
 This library is built with [SBT](http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html), which is automatically downloaded by the included shell script. To build a JAR file simply run `sbt/sbt assembly` from the project root.
-
--- a/README.md
+++ b/README.md
@@ -14,12 +14,12 @@ You can link against this library in your program at the following coordiates:
 ```
 groupId: com.databricks
 artifactId: spark-csv_2.10
-version: 0.2.0
+version: 1.0.0
 ```
 The spark-csv assembly jar file can also be added to a Spark using the `--jars` command line option.  For example, to include it when starting the spark shell:
 
 ```
-$ bin/spark-shell --jars spark-csv-assembly-0.2.0.jar
+$ bin/spark-shell --jars spark-csv-assembly-1.0.0.jar
 ```
 
 ## Features
@@ -67,6 +67,10 @@ import com.databricks.spark.csv.JavaCsvParser;
 
 DataFrame cars = (new JavaCsvParser()).withUseHeader(true).csvFile(sqlContext, "cars.csv");
 ```
+
+### Saving as CSV
+You can save your DataFrame using `saveAsCsvFile` function. The function allows you to specify the delimiter and whether we should generate a header row for the table (each header has name `C$i` where `$i` is column index). For example:
+```myDataFrame.saveAsCsvFile("/mydir", Map("delimiter" -> "|", "header" -> "true"))```
 
 ## Building From Source
 This library is built with [SBT](http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html), which is automatically downloaded by the included shell script. To build a JAR file simply run `sbt/sbt assembly` from the project root.
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -15,32 +15,34 @@
  */
 package com.databricks.spark.csv
 
-import org.apache.hadoop.fs.{FileSystem, Path}
-import org.slf4j.LoggerFactory
-
-import org.apache.commons.csv._
-
-import org.apache.spark.sql._
-import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.sources.TableScan
-import org.apache.spark.sql.types.{StructType, StructField, StringType}
+import java.io.IOException
 
 import scala.collection.JavaConversions._
 import scala.util.control.NonFatal
+
+import org.apache.commons.csv._
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql._
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.sources.{BaseRelation, InsertableRelation, TableScan}
+import org.apache.spark.sql.types.{StructType, StructField, StringType}
+import org.slf4j.LoggerFactory
+
 
 case class CsvRelation protected[spark] (
     location: String,
     useHeader: Boolean,
     delimiter: Char,
     quote: Char,
-    userSchema: StructType = null)(@transient val sqlContext: SQLContext) extends TableScan {
+    userSchema: StructType = null)(@transient val sqlContext: SQLContext)
+  extends BaseRelation with TableScan with InsertableRelation {
 
   private val logger = LoggerFactory.getLogger(CsvRelation.getClass)
 
-  val schema = inferSchema()
+  var schema = inferSchema()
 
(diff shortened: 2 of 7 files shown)


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -15,32 +15,34 @@
...
+  extends BaseRelation with TableScan with InsertableRelation {
 
   private val logger = LoggerFactory.getLogger(CsvRelation.getClass)
 
-  val schema = inferSchema()
+  var schema = inferSchema()
 
   // By making this a lazy val we keep the RDD around, amortizing the cost of locating splits.
-  lazy val buildScan = {
+  def buildScan = {
```
vlyubin: If you change `lazy val` to `def`, `INSERT OVERWRITE` works. Otherwise the problem that I brought up occurs.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -43,21 +43,23 @@ package object csv {
...
-          var firstRow: Boolean = true
+          var firstRow: Boolean = generateHeader
 
           override def hasNext = iter.hasNext
 
           override def next: String = {
             if (firstRow) {
               firstRow = false
-              header + "\n" + iter.next.mkString(",")
+              header + "\n" + iter.next.mkString(delimiter)
```
vlyubin: Note that I had to change the logic slightly (so that "INSERT OVERWRITE WORKS"). In particular, I have to use the same delimiter as before.
Also, the default logic was to always add a header row, while it might be the case that we already have them. Thus the default behavior should be to write the RDD as is, and if the user who is using this functions wants us to add a header, we will pass `generateHeader ` in parameters (I'll add this to README).

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -15,32 +15,34 @@
...
     delimiter: Char,
     quote: Char,
-    userSchema: StructType = null)(@transient val sqlContext: SQLContext) extends TableScan {
+    userSchema: StructType = null)(@transient val sqlContext: SQLContext)
+  extends BaseRelation with TableScan with InsertableRelation {
 
   private val logger = LoggerFactory.getLogger(CsvRelation.getClass)
 
-  val schema = inferSchema()
+  var schema = inferSchema()
```
falaki: Please keep it a val.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +137,29 @@ case class CsvRelation protected[spark] (
...
+  // The function below was borrowed from JSONRelation
+  override def insert(data: DataFrame, overwrite: Boolean) = {
+    val filesystemPath = new Path(location)
+    val fs = filesystemPath.getFileSystem(sqlContext.sparkContext.hadoopConfiguration)
+
+    if (overwrite) {
+      try {
+        fs.delete(filesystemPath, true)
+      } catch {
+        case e: IOException =>
```
falaki: Shouldn't we fail if this exception happens? If we catch the exception and continue with saving the data, wouldn't we end up with inconsistent data?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +137,29 @@ case class CsvRelation protected[spark] (
...
+          throw new IOException(
+            s"Unable to clear output directory ${filesystemPath.toString} prior"
+              + s" to INSERT OVERWRITE a CSV table:\n${e.toString}")
+      }
+      // Write the data.
+      data.saveAsCsvFile(location, Map("delimiter" -> delimiter.toString))
+      // Right now, we assume that the schema is not changed. We will not update the schema.
+      // schema = data.schema
```
falaki: Please remove.

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -109,4 +157,32 @@ class CsvSuite extends FunSuite {
     assert(cars.schema.fields(0).name == "C0")
     assert(cars.schema.fields(2).name == "C2")
   }
+
+  test("INSERT OVERWRITE test") {
```
falaki: Please follow the naming convention of other tests: SQL test insert overwrite

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -43,21 +43,23 @@ package object csv {
   }
 
   implicit class CsvSchemaRDD(dataFrame: DataFrame) {
-    def saveAsCsvFile(path: String): Unit = {
+    def saveAsCsvFile(path: String, parameters: Map[String, String] = Map()): Unit = {
       // TODO(hossein): For nested types, we may want to perform special work
-      val header = dataFrame.columns.map(c => s""""$c"""").mkString(",")
+      val delimiter = parameters.getOrElse("delimiter", ",")
+      val generateHeader = parameters.getOrElse("generateHeader", "false").toBoolean
```
falaki: I think simply header is better than "generateHeader."

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -43,21 +43,23 @@ package object csv {
...
+      val header = dataFrame.columns.map(c => s""""$c"""").mkString(delimiter)
```
falaki: Can we avoid constructing this sting if header is false?

src/main/scala/com/databricks/spark/csv/package.scala
falaki: Make sense. Please update the README

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -26,6 +28,52 @@ class CsvSuite extends FunSuite {
...
   val carsAltFile = "src/test/resources/cars-alternative.csv"
   val emptyFile = "src/test/resources/empty.csv"
+  val tempEmptyDir = "target/test/empty/"
+
+  /**
+   * This function deletes a file or a directory with everything that's in it. This function is
+   * copied from Spark with minor modifications made to it. See original source at:
+   * github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/Utils.scala
+   */
+  private def deleteRecursively(file: File) {
```
falaki: Please move this to a test utility file.

falaki:
@vlyubin thanks for doing this. I left some comments.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
vlyubin: Oh, whoops. Originally I had `schema = data.schema` on, so I needed var for that.

falaki:
@vlyubin would you please update this?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
vlyubin: Wait, it does fail, doesn't it? It catches exception and throws one with a nice message. Or if it's unmatched (unlikely), we will get an even uglier exception :)
Anyway, that's the way JSON does it.

src/main/scala/com/databricks/spark/csv/package.scala
vlyubin: Note that in other cases `header` is referring to wether first row is header. Here it means whether we should generate a header for the file we output. Naming this `header` again might be confusing. Though that's up to you to decide. If you still thing `header` is good, I'll make the change.

src/main/scala/com/databricks/spark/csv/package.scala
vlyubin: Will do.

src/main/scala/com/databricks/spark/csv/package.scala
vlyubin: We can but it will be a bit hacky. We can't move this line inside mapPartitions (it uses `dataFrame`), but I can use an if here, and leave it empty if `header` is false. I guess this is tolerable, I'll make the change.

README.md
```diff
@@ -14,12 +14,12 @@ You can link against this library in your program at the following coordiates:
 ```
 groupId: com.databricks
 artifactId: spark-csv_2.10
-version: 0.2.0
+version: 1.0.0
```
vlyubin: Per Michael and Aaron, we will make csv and avro versions that work with spark 1.3+ start with 1.0.0

src/main/scala/com/databricks/spark/csv/package.scala
falaki: I think header is more concise and clear in this context.

falaki:
Thanks @vlyubin. Merging this.

PR #30, opened 2015-03-19
Use the same library when Reading and Writing CSV

abendada:
- Added Escape Character
- Use CSVPrinter when writing CSV Files

### DIFF
--- a/src/main/java/com/databricks/spark/csv/JavaCsvParser.java
+++ b/src/main/java/com/databricks/spark/csv/JavaCsvParser.java
@@ -27,6 +27,7 @@ public class JavaCsvParser {
   private Boolean useHeader = true;
   private Character delimiter = ',';
   private Character quote = '"';
+  private Character escape = '\\';
   private StructType schema = null;
 
   public JavaCsvParser withUseHeader(Boolean flag) {
@@ -44,6 +45,11 @@ public class JavaCsvParser {
     return this;
   }
 
+  public JavaCsvParser withEscapeChar(Character escape) {
+    this.escape = escape;
+    return this;
+  }
+
   public JavaCsvParser withSchema(StructType schema) {
     this.schema = schema;
     return this;
@@ -52,7 +58,7 @@ public class JavaCsvParser {
   /** Returns a Schema RDD for the given CSV path. */
   public DataFrame csvFile(SQLContext sqlContext, String path) {
     CsvRelation relation = new
-            CsvRelation(path, useHeader, delimiter, quote, schema, sqlContext);
+            CsvRelation(path, useHeader, delimiter, quote, escape, schema, sqlContext);
     return sqlContext.baseRelationToDataFrame(relation);
   }
 }
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -26,6 +26,7 @@ class CsvParser {
   private var useHeader: Boolean = true
   private var delimiter: Character = ','
   private var quote: Character = '"'
+  private var escape: Character = '\\'
   private var schema: StructType = null
 
   def withUseHeader(flag: Boolean): CsvParser = {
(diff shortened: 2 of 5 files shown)


falaki:
@abendada Would you add tests for this and make it pass?

abendada:
I added a test for escape
The test consist of:
- reading a csv file with "\" as escape character
- check the parsed content
- save the dataFrame as a temp file without providing an escape character the default (") is used if necessary
- check that temp file is well formed
- read the temp file and check that the content well parsed

Note: quote and escape should have Character as type so nullable.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -33,7 +33,8 @@ case class CsvRelation protected[spark] (
     location: String,
     useHeader: Boolean,
     delimiter: Char,
-    quote: Char,
+    quote: Character,
```
falaki: Use Char

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -33,7 +33,8 @@ case class CsvRelation protected[spark] (
...
+    escape: Character,
```
falaki: Use Char

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -45,28 +54,29 @@ package object csv {
...
+      val strRDD = dataFrame.rdd.mapPartitions { iter =>
+        var firstRow: Boolean = generateHeader
+        val newIter = iter.map(_.toSeq.toArray)
+        val stringWriter = new StringWriter()
+        val csvPrinter = new CSVPrinter(stringWriter, csvFileFormat)
+        if (firstRow) {
+          firstRow = false
+          csvPrinter.printRecord(header:_*)
         }
+        csvPrinter.printRecords(WrapAsJava.asJavaIterable(newIter.toIterable))
```
falaki: This can be inefficient because it seems to me we are traversing the iterator multiple times for each partition. I know CSVPrinter API is not very flexible here, but can we avoid multiple traversals?

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -45,28 +54,29 @@ package object csv {
...
-            } else {
-              iter.next.mkString(delimiter)
-            }
-          }
+      val strRDD = dataFrame.rdd.mapPartitions { iter =>
+        var firstRow: Boolean = generateHeader
+        val newIter = iter.map(_.toSeq.toArray)
+        val stringWriter = new StringWriter()
+        val csvPrinter = new CSVPrinter(stringWriter, csvFileFormat)
+        if (firstRow) {
```
falaki: this condition is always true for each partition. Note that previously we were making the check inside the Iterator.next().

src/main/scala/com/databricks/spark/csv/CsvParser.scala
```diff
@@ -50,7 +56,7 @@ class CsvParser {
 
   /** Returns a Schema RDD for the given CSV path. */
   def csvFile(sqlContext: SQLContext, path: String): DataFrame = {
-    val relation: CsvRelation = CsvRelation(path, useHeader, delimiter, quote, schema)(sqlContext)
+    val relation: CsvRelation = CsvRelation(path, useHeader, delimiter, quote, escape, schema)(sqlContext)
```
falaki: This is more than 100 characters.

src/main/java/com/databricks/spark/csv/JavaCsvParser.java
```diff
@@ -44,6 +45,11 @@ public JavaCsvParser withQuoteChar(Character quote) {
     return this;
   }
 
+  public JavaCsvParser withEscapeChar(Character escape) {
```
falaki: This file is removed in master. No need for these changes.

falaki:
@abendada left a few comments. Would you please also update the README for the new escape character option, and rebase this? Thanks!

PR #31, opened 2015-03-25
Add support for specifying a compression codec when saving

pashields:
Didn't see any existing tests to update. Did I miss anything?

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/package.scala
+++ b/src/main/scala/com/databricks/spark/csv/package.scala
@@ -16,6 +16,7 @@
 package com.databricks.spark
 
 import org.apache.spark.sql.{SQLContext, DataFrame}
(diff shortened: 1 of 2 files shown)


falaki:
Thanks @pashields. Two suggestions:
1. Add tests for this
2. I think compression can be an option to save(). If you are interested look into ```CreatableRelationProvider``` in Spark 1.3 and add support for that?

falaki:
Thanks @pashields this looks good. Would you please rebase it? I am going to cut v1.0.1 with this.

pashields:
Yeah, I will hide my shame with a rebase. I'm still looking at adding support in ```save()```, but I'm new to scala and spark so let's not hold this up for that.

I'll push in a minute.

pashields:
Ok, should be good @falaki. Thanks for your help with the PR.

PR #33, opened 2015-03-25
Scala 2.11

falaki:
* We don't need JavaCsvParser any more
* Updated documentation with instructions for load/save
* Bumped up version of Scala to 2.11 (still compiling against 2.10)


dhm116:
:+1:

dhm116:
`HashMap<String, String> options = new HashMap<>();` isn't valid in java 1.6 (which is what Travis appears to be building with) - maybe replace with `HashMap<String, String> options = new HashMap<String, String>();` to be safe?

README.md
```diff
@@ -23,24 +23,19 @@ $ bin/spark-shell --jars spark-csv-assembly-1.0.0.jar
...
 
 ## Features
-These examples use a CSV file available for download [here](https://github.com/databricks/spark-csv/raw/master/src/test/resources/cars.csv):
-
-```
-$ wget https://github.com/databricks/spark-csv/raw/master/src/test/resources/cars.csv
-```
+This package allows reading CSV files in local or distributed filesystem as [Spark DataFrames](https://spark.apache.org/docs/1.3.0/sql-programming-guide.html).
+When reading files the API accepts several options:
+* path: location of files. Similar to Spark can be a wildcard.
```
marmbrus: Similar to Spark can accept standard hadoop globbing expressions.

README.md
```diff
@@ -23,24 +23,19 @@ $ bin/spark-shell --jars spark-csv-assembly-1.0.0.jar
...
+* header: when set to true the first line of files will be used to name columns and will not be included in data. All types will be assumed string.
```
marmbrus: Maybe specify default value (i assume false?)

README.md
```diff
@@ -23,24 +23,19 @@ $ bin/spark-shell --jars spark-csv-assembly-1.0.0.jar
...
+* delimiter: by default lines are delimited using ',', but delimiter can be set to any character
+* quote: by default the quote character is '"', but can be set to any character. Delimiters inside quotes are ignored
 
-### Scala API
+The package also support saving simple (non-nested) DataFrame. When saving you can specify the delimiter and whether we should generate a header row for the table (each header has name `c$i` where `$i` is column index). See following examples for more details.
```
marmbrus: > each header has name `c$i` where `$i` is column index

What does this mean?  Does it not just use the column names?

README.md
```diff
@@ -59,18 +54,64 @@ USING com.databricks.spark.csv
...
 ```
 
+### Scala API
+The recommended way to load CSV data is using the load/save functions in SQLContext.
+
+```scala
+import org.apache.spark.sql.SQLContext
+
+val sqlContext = new SQLContext(sc)
+val df = sqlContext.load("com.databricks.spark.csv", Map("path" -> "cars.csv", "header" -> "true"))
```
marmbrus: You might consider the simple example too:
```
sqlContext.load("cars.csv", "com.databricks.spark.csv")
```

README.md
```diff
@@ -59,18 +54,64 @@ USING com.databricks.spark.csv
...
+df.select("year", "model").save("newcars.csv", Map("header" -> "false", "delimiter" -> "\t"))
```
marmbrus: need to specify the source here i think.

README.md
falaki: Default is true, but I can change to false if you think that is a better choice.

README.md
```diff
@@ -59,18 +54,64 @@ USING com.databricks.spark.csv
...
-```myDataFrame.saveAsCsvFile("/mydir", Map("delimiter" -> "|", "header" -> "true"))```
+### Python API
+In Python you can read and save CSV files using load/save functions.
+
+```python
+from pyspark.sql import SQLContext
+sqlContext = SQLContext(sc)
+
+df = sqlContext.load("com.databricks.spark.csv", path = "cars.csv", header = True)
+df.select("year", "model").save("newcars.csv", header = False, delimiter = "\t")
```
marmbrus: need the source here too.

README.md
```diff
@@ -59,18 +54,64 @@ USING com.databricks.spark.csv
...
 
-DataFrame cars = (new JavaCsvParser()).withUseHeader(true).csvFile(sqlContext, "cars.csv");
+SQLContext sqlContext = new SQLContext(sc);
+
+HashMap<String, String> options = new HashMap<String, String>();
+options.put("header", "true");
+options.put("path", "cars.csv");
+
+DataFrame df = sqlContext.load("com.databricks.spark.csv", options);
+df.select("name", "age").save("newcars.csv", options);
```
marmbrus: need the source here

marmbrus:
I also don't see a test for the `save` code path.  Might be good to add one.

Otherwise LGTM

README.md
marmbrus: Oh, I think false is better.  Possibly dropping data is more confusing that having them see their headers in the results.

PR #37, opened 2015-03-28
Be permissive of rows that have fewer tokens than the schema

### LINKED ISSUE #11: Schema validation
The parser crashes and there's no way to catch the exception if the schema provided has more fields than the actual CSV file.

val parser = new CsvParser().withDelimiter(delimiter).withQuoteChar(quote).withUseHeader(useHeader)
parser.withSchema(myCustomSchema)

The parser keeps on looping on the entire file, and throws ArrayIndexOutOfBoundsException for each line in the file.

falaki:
Adds a new option for parsing mode that determines the parsing mode. By default it is PERMISSIVE. Possible values are: 
* PERMISSIVE: tries to parse all lines. When a line has fewer tokens than the schema, we produce a row with nulls for remaining tokens.
* DROPMALFORMED: drops lines which have fewer or more tokens than expected
* FAILFAST: aborts with a RuntimeException if encounters any malformed line

Closes #11


falaki:
@aarondav please take a look when you get a chance.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -28,17 +28,27 @@ import org.apache.spark.sql.sources.{BaseRelation, InsertableRelation, TableScan
...
     quote: Char,
+    parseMode: String,
     userSchema: StructType = null)(@transient val sqlContext: SQLContext)
   extends BaseRelation with TableScan with InsertableRelation {
 
   private val logger = LoggerFactory.getLogger(CsvRelation.getClass)
 
+  // Parse mode flags
+  if (!ParseModes.isValidMode(parseMode)) {
+    logger.warn(s"$parseMode is not a valid parse mode. Using ${ParseModes.DEFAULT}.")
```
aarondav: This is not true, I believe, given the logic below; all three would be false. Maybe defined permissive = (!failFast || !dropMalformed)?

README.md
```diff
@@ -31,6 +31,10 @@ When reading files the API accepts several options:
 * header: when set to true the first line of files will be used to name columns and will not be included in data. All types will be assumed string. Default value is false.
 * delimiter: by default lines are delimited using ',', but delimiter can be set to any character
 * quote: by default the quote character is '"', but can be set to any character. Delimiters inside quotes are ignored
+* mode: determines the parsing mode. By default it is PERMISSIVE. Possible values are:
+  * PERMISSIVE: tries to parse all lines
```
aarondav: mention that nulls are inserted for missing columns

aarondav:
LGTM, only minor comments.

Issue #38, opened 2015-03-28
java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String


MarkRijckenberg:
Hi,

I have written the following procedure and followed it:

https://mark911.wordpress.com/2015/03/21/how-to-build-pysparkmaven-and-apache-spark-from-github-sources-using-a-bash-script-in-ubuntu-14-04-lts-64-bit/

Importing the csv file via the procedure I wrote, works when using com.databricks:spark-csv version 2.10:1.0.0, but does not work using 2.11:1.0.1. When using 2.11:1.0.1, the Excel header line gets stripped and I need to use the C1 , C2 column names instead of the proper column names in the Excel file.

Is this a new bug in 2.11:1.0.1 ??

Regards,

Mark

falaki:
In version 1.0.1 we changed the default value of header flag from true to false. So if you have a header line in your file, you need to dome something like:
```python
df = sqlContext.load("com.databricks.spark.csv", path = "cars.csv", header = True)
```

MarkRijckenberg:
Thanks for your answer. However, I previously already tried adding header = True and the command only gives an error:

>>> df = sqlContext.load("com.databricks.spark.csv", path = "cars.csv", header = True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: load() got multiple values for keyword argument 'path'
>>> 

Using path =  without adding header = True also gives an error:

>>> df = sqlContext.load("com.databricks.spark.csv", path = "cars.csv")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: load() got multiple values for keyword argument 'path'
>>> 

Here is what I get when I form the command my way and add "header = True" ( error is (py4j.protocol.Py4JJavaError: An error occurred while calling o21.load.
: java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String ) )

>>> df = sqlContext.load("cars.csv", "com.databricks.spark.csv", header = True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ulysses/spark/python/pyspark/sql/context.py", line 482, in load
    df = self._ssql_ctx.load(source, joptions)
  File "/home/ulysses/spark/python/lib/py4j-0.8.2.1-src.zip/py4j/java_gateway.py", line 538, in __call__
[...]

PR #39, opened 2015-03-28
Update README.md

MarkRijckenberg:
Reason for pull request:  see   https://github.com/databricks/spark-csv/issues/38

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ In Python you can read and save CSV files using load/save functions.
 from pyspark.sql import SQLContext
 sqlContext = SQLContext(sc)
 
-df = sqlContext.load("com.databricks.spark.csv", path = "cars.csv", header = True)
+df = sqlContext.load(source="com.databricks.spark.csv", header="true",path = "cars.csv")
 df.select("year", "model").save("newcars.csv", "com.databricks.spark.csv")
 ```
 


falaki:
Thanks!

rxin:
Maybe Spark's data source API should convert common Python types back into proper Strings. 

@davies can we add that?

rxin:
I filed https://issues.apache.org/jira/browse/SPARK-6591 for this.

MarkRijckenberg:
Thanks for filing the bug report at https://issues.apache.org/jira/browse/SPARK-6591

However, please keep in mind that the whole issue is a little broader than only converting   header = True  to header="true"

If you follow my procedure here:

https://mark911.wordpress.com/2015/03/21/how-to-build-pysparkmaven-and-apache-spark-from-github-sources-using-a-bash-script-in-ubuntu-14-04-lts-64-bit/

and you leave out source= or header= or leave out path = or if you use header="True" or any combination of these , the command will still not work. So the use of header="true" is case sensitive in Ubuntu. And the use of source=  , header= and path= is apparently required...

The rules to getting

df = sqlContext.load(source="com.databricks.spark.csv", header="true",path = "cars.csv")

to work are stricter - at least in Ubuntu -  than just converting the Python type back into Strings.

rxin:
I was not suggesting the Spark load API will fix this. I was merely saying we should make it more convenient for users.

PR #42, opened 2015-03-31
Add support for quoting on save.

pashields:
Adds support for saving with quoting around each elements and any existing occurrences of the quote being escaped. Because ```CSVFormat``` can handle unquoting, the roundtrip is lossless.

A few notes/questions:
* Performance on this is not that great. I clocked some example data taking about 1.5x as long to save quoted vs. unquoted. There should be little cost to non-quoted saves with this patch (just an additional function call or two of indirection). Any ideas for performance improvements would be great!
* I brought in ```commons-lang```. Spark already requires this, so hopefully it's not too big of a deal. I can't decide if it's better to have an explicit version dep or go with provided as such.
* I attempted to use ```commons-csv```, but was hampered by the insistence on [writing the delimiter at the top of every output](https://github.com/apache/commons-csv/blob/trunk/src/main/java/org/apache/commons/csv/CSVPrinter.java#L127). It might not be too hard to undo this, but like a more kludge-y approach.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,8 +12,6 @@ libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.3.0" % "provided"
 
 libraryDependencies += "org.apache.commons" % "commons-csv" % "1.1"
 
-libraryDependencies += "org.apache.commons" % "commons-lang3" % "3.3.2"
-
 libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.5" % "provided"
 
 libraryDependencies += "com.novocode" % "junit-interface" % "0.9" % "test"
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,8 @@ libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.3.0" % "provided"
 
 libraryDependencies += "org.apache.commons" % "commons-csv" % "1.1"
 
+libraryDependencies += "org.apache.commons" % "commons-lang3" % "3.3.2"
+
 libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.5" % "provided"
 
 libraryDependencies += "com.novocode" % "junit-interface" % "0.9" % "test"
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -57,6 +57,11 @@ class CsvParser {
     this
   }
 
+  def withEscape(escapeChar: Character): CsvParser = {
+    this.escape = escapeChar
+    this
+  }
+
   /** Returns a Schema RDD for the given CSV path. */
   @throws[RuntimeException]
   def csvFile(sqlContext: SQLContext, path: String): DataFrame = {
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -28,6 +28,7 @@ class CsvParser {
   private var useHeader: Boolean = false
   private var delimiter: Character = ','
   private var quote: Character = '"'
+  private var escape: Character = '\\'
   private var schema: StructType = null
   private var parseMode: String = ParseModes.DEFAULT
 
@@ -64,6 +65,7 @@ class CsvParser {
       useHeader,
       delimiter,
       quote,
+      escape,
       parseMode,
       schema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -63,6 +63,7 @@ case class CsvRelation protected[spark] (
     val csvFormat = CSVFormat.DEFAULT
       .withDelimiter(delimiter)
       .withQuote(quote)
+      .withEscape('\\')
       .withSkipHeaderRecord(false)
       .withHeader(fieldNames: _*)
 
@@ -88,6 +89,7 @@ case class CsvRelation protected[spark] (
       val csvFormat = CSVFormat.DEFAULT
         .withDelimiter(delimiter)
         .withQuote(quote)
+        .withEscape('\\')
         .withSkipHeaderRecord(false)
       val firstRow = CSVParser.parse(firstLine, csvFormat).getRecords.head.toList
       val header = if (useHeader) {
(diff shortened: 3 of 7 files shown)


src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -49,6 +50,28 @@ package object csv {
     }
   }
 
+  implicit class CsvRow(row: Row) {
```
falaki: These methods would become exposed as part of our public interface. We don't want that, especially since they are fairly low-level.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -58,6 +81,7 @@ package object csv {
                       compressionCodec: Class[_ <: CompressionCodec] = null): Unit = {
       // TODO(hossein): For nested types, we may want to perform special work
       val delimiter = parameters.getOrElse("delimiter", ",")
+      val quote = parameters.get("quote")
```
falaki: This would throw an exception is user did not specify quote. We should have a default.

build.sbt
```diff
@@ -12,6 +12,8 @@ libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.3.0" % "provided"
 
 libraryDependencies += "org.apache.commons" % "commons-csv" % "1.1"
 
+libraryDependencies += "org.apache.commons" % "commons-lang3" % "3.3.2"
```
falaki: I think marking it as provided is a good idea.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -63,6 +63,7 @@ case class CsvRelation protected[spark] (
     val csvFormat = CSVFormat.DEFAULT
       .withDelimiter(delimiter)
       .withQuote(quote)
+      .withEscape('\\')
```
falaki: It would be nice not to hardcode the escape character, and get it from options.

pashields:
Just pushed an update that actually does use the apache CSVFormatter. Apologies for the runaround. I figured out how to get what we need out of it. I was tempted to expose [QuoteMode](https://commons.apache.org/proper/commons-csv/archives/1.1/apidocs/org/apache/commons/csv/QuoteMode.html) but didn't think the API should be tied to the ```commons-csv``` one.

I'll note individually, but the other issues should be addressed as well.

src/main/scala/com/databricks/spark/csv/package.scala
pashields: Removed.

src/main/scala/com/databricks/spark/csv/package.scala
pashields: Is there some reason this won't be a scala map? Shouldn't it return an option we can match on?

```
scala> val m: Map[String, String] = Map()
m: Map[String,String] = Map()

scala> val noString = m.get("key")
noString: Option[String] = None
```

build.sbt
pashields: dep removed.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
pashields: Now configurable.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -58,6 +63,20 @@ package object csv {
...
+      }
+
+      val escape = parameters.getOrElse("escape", "\\")
+      val escapeChar = if (escape.length == 1) {
+        escape.charAt(0)
+      } else {
+        throw new Exception("Escape character cannot be more than one character.")
+      }
+
+      val quote = parameters.get("quote")
```
falaki: Would you use ```getOrElse``` here as well?
Also I think this is a better place to make sure quote is a string of length one, and convert it to a Char.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -65,17 +84,35 @@ package object csv {
...
       val strRDD = dataFrame.rdd.mapPartitions { iter =>
+        val csvFormatBase = CSVFormat.DEFAULT
+          .withDelimiter(delimiterChar)
+          .withEscape(escapeChar)
+          .withSkipHeaderRecord(false)
+          .withNullString("null")
+
+        val csvFormat = quote match {
+          case Some(s) => {
+            if (s.length == 1) {
```
falaki: This check is better done outside of the loop, otherwise we will throw the exception on every element.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -15,8 +15,10 @@
  */
 package com.databricks.spark
 
-import org.apache.spark.sql.{SQLContext, DataFrame}
+import org.apache.commons.csv.CSVFormat
+import org.apache.spark.sql.{SQLContext, DataFrame, Row}
 import org.apache.hadoop.io.compress.CompressionCodec
+import org.apache.commons.lang3.StringUtils
```
falaki: Nit: it is better to sort the imports according to Spark's code style guide: https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide#SparkCodeStyleGuide-Imports

src/main/scala/com/databricks/spark/csv/CsvParser.scala
```diff
@@ -64,6 +65,7 @@ class CsvParser {
       useHeader,
       delimiter,
       quote,
+      escape,
```
falaki: Would you also add a ```with escape(escapeChar: Character)``` to this?

falaki:
@pashields thanks for the changes. This looks much better. I am going to cut a 1.0.2 version, and will include this one.

src/main/scala/com/databricks/spark/csv/package.scala
falaki: I see. So if quote is provided, you would quote all fields, and otherwise not? That makes sense.

src/main/scala/com/databricks/spark/csv/package.scala
falaki: Just saw your other comment. We could use None as a signal for not quoting. But I still think it is better to check the length here instead of inside the iterator.

pashields:
Pushed new changes. Should address your points. Really good point about pulling that string length check out of the loop.

And yeah, the idea is to use the option to give the user choice about quoting on save. I can imagine that we might want escaping to be optional as well, make these optional on reads too, but that can be a separate PR if someone wants it.

I'll be traveling today, so I may be slow to answer emails, but I'll be able to make last minute fixes and squash later today. Thanks for the help with the PR!

falaki:
Thanks! This looks good.

PR #52, opened 2015-04-19
Support loading of RDDs (of case classes) from CSV.

rayortigas:
I'm still in RDD-land, so I'd like something like this to avoid writing things like 

```scala
val rdd = sqlContext.csvFile(path, useHeader = false).map { row =>
  Foo(row.getString(0).toInt, row.getString(1).toInt, row.getString(2).toDouble)
}
```

So instead we can write

```scala
val rdd = sqlContext.csvFileToRDD[Foo](path, useHeader = false)
```

I tried to be minimally invasive here by building on top of `csvFile`. With more refactoring, I probably would've teased out some stuff in `CsvRelation`, but I hope this PR is useful in its present form.

Regards,
Ray

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -60,3 +60,7 @@ sparkComponents += "sql"
 libraryDependencies += "org.scalatest" %% "scalatest" % "2.2.1" % "test"
 
 libraryDependencies += "com.novocode" % "junit-interface" % "0.9" % "test"
+
+// Fork to help tests of methods using reflection.
+// See https://issues.apache.org/jira/browse/SPARK-5281.
+fork in Test := true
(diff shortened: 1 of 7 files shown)


rxin:
@rayortigas   this seems like something that can easily live outside of the CSV package.  There isn't anything specific to CSV about this one. 

As a matter of fact it probably deserves to either be part of the DataFrame API, or just an implicit conversion on DataFrame to add the following:

```scala
// or called toTyped, or typedRDD
def toTypedRDD[T : scala.reflect.runtime.universe.TypeTag : scala.reflect.ClassTag]: RDD[T] = {
   ...
}
```

rayortigas:
@rxin I'd love for DataFrames to support it directly... I picked CSV first because the conversion was more straightforward (just a row of primitives). :D

Maybe I'll put together a PR for spark proper that handles more complex objects? I see what ScalaReflection is doing (and I think I saw the latest refactoring), so I'll take a cue from that.

rayortigas:
OK, I opened https://github.com/apache/spark/pull/5713. Thanks for the suggestion @rxin!

PR #53, opened 2015-04-21
Bumping up version of spark packages plugin

falaki:
Using latest version of spark packages sbt plugin (0.2.1)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,8 +10,6 @@ parallelExecution in Test := false
 
 crossScalaVersions := Seq("2.10.4", "2.11.6")
 
-libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.3.0" % "provided"
-
 libraryDependencies += "org.apache.commons" % "commons-csv" % "1.1"
 
 libraryDependencies += "com.univocity" % "univocity-parsers" % "1.5.1"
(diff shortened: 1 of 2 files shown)


brkyvz:
@falaki 
Could you please update the following in the build file?
 - `sparkPackageName` -> `spName` // sparkPackageName is deprecated
 - Remove [this line](https://github.com/databricks/spark-csv/blob/464a3e09529d97cdbb7f74b3d1bb5767c3c27ca2/build.sbt#L11). This is handled [here](https://github.com/databricks/spark-csv/blob/464a3e09529d97cdbb7f74b3d1bb5767c3c27ca2/build.sbt#L58)
 - add the line: `spAppendScalaVersion := true`

Thanks!

falaki:
@brkyvz Thanks! Did we fix the bug in spark-packages.org with respect o appended scala version?

brkyvz:
In Review. I'll ping Xiangrui tonight

brkyvz:
@falaki the bug in spark-packages.org should be fixed now

falaki:
@brkyvz Would you please take a look.

brkyvz:
@falaki You can also include the option
`spIncludeMaven := true`
to include the original Maven coordinate (`com.databricks:spark-csv_2.11:v`) when publishing to Spark Packages. Then, you will need to publish to Maven Central first, before publishing to Spark Packages.

PR #67, opened 2015-06-02
[Issue 65] Parsing CSV with '\t' in notebook fails

### LINKED ISSUE #65: Parsing CSV with '\t' in notebook fails
```sql
CREATE TABLE tbl_activity_g0_t60_raw
USING com.databricks.spark.csv
options (path "data", header "false", delimiter '\t')
```
Will fail with:
```
"Delimiter cannot be more than one character."
```

falaki:
We use apache commons-lang to unescape the delimiter string. Also added a unit test for this.

closes #65

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
@@ -15,6 +15,7 @@
  */
 package com.databricks.spark.csv
 
+import org.apache.commons.lang3.StringEscapeUtils
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, SaveMode, SQLContext}
 import org.apache.spark.sql.sources._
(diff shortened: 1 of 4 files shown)


rxin:
Do we really need an extra library for this?

falaki:
Not really, have a better standard way of unescaping strings? I am open to using anything else.

brkyvz:
I don't know a better way of unescaping strings, therefore I'm going to LGTM.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -49,7 +50,7 @@ class DefaultSource
       parameters: Map[String, String],
       schema: StructType) = {
     val path = checkPath(parameters)
-    val delimiter = parameters.getOrElse("delimiter", ",")
+    val delimiter = StringEscapeUtils.unescapeJava(parameters.getOrElse("delimiter", ","))
```
rxin: can't we just replace "\t" string literal with a tab here? I know there are other things, but I don't think we should handle everything anyway (like \n)

falaki: I prefer the more general approach. There are some crazy delimiters out there in the wild!

rxin: What happens if "\n" is specified as the delimiter? Is this library going to crash Spark?

Issue #64, opened 2015-05-26
[Spark 1.4] ClassCastException: java.lang.String cannot be cast to org.apache.spark.sql.types.UTF8String


evacchi:
Currently broken with Spark 1.4 when an explicit schema with non-string fields is specified; e.g.:

```scala
val schema = StructType(Seq(StructField("id", LongType, false)))
```

falaki:
Please provide additional details including sample of data and the code that leads to this error.

rxin:
This is probably because we don't do inbound conversion in Spark itself. cc @yhuai

pwendell:
@yhuai do you mind creating a JIRA to track this - it could be release blocker (/cc @marmbrus)

yhuai:
I see. The issue is that Spark-CSV is using the Spark SQL's internal API to do the type cast, which assumes values are stored in Spark SQL's internal representation (e.g. `UTF8String`). We will fix it in Spark-CSV.


btw, here is the stack trace.

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 31.0 failed 4 times, most recent failure: Lost task 0.3 in stage 31.0 (TID 41, 192.168.1.26): java.lang.ClassCastException: java.lang.String cannot be cast to org.apache.spark.sql.types.UTF8String
	at org.apache.spark.sql.catalyst.expressions.Cast$$anonfun$castToLong$1$$anonfun$apply$28.apply(Cast.scala:249)
	at org.apache.spark.sql.catalyst.expressions.Cast.org$apache$spark$sql$catalyst$expressions$Cast$$buildCast(Cast.scala:111)
	at org.apache.spark.sql.catalyst.expressions.Cast$$anonfun$castToLong$1.apply(Cast.scala:249)
	at org.apache.spark.sql.catalyst.expressions.Cast.eval(Cast.scala:436)
	at org.apache.spark.sql.catalyst.expressions.InterpretedMutableProjection.apply(Projection.scala:68)
	at org.apache.spark.sql.catalyst.expressions.InterpretedMutableProjection.apply(Projection.scala:52)
	at com.databricks.spark.csv.CsvRelation$$anonfun$com$databricks$spark$csv$CsvRelation$$parseCSV$1.apply(CsvRelation.scala:150)
	at com.databricks.spark.csv.CsvRelation$$anonfun$com$databricks$spark$csv$CsvRelation$$parseCSV$1.apply(CsvRelation.scala:130)
[...]

PR #68, opened 2015-06-03
[Issue 64] Do not use internal SparkSQL type conversion

falaki:
We were relying on utility classes under ```org.apache.spark.sql.catalyst.expressions._```. This patch removes all those dependancies.

cc @pwendell and @rxin 

closes issue #64


yhuai:
I guess creating those converters as functions first and then apply them to the RDD of rows will give us better performance (like what do we at [here] (https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/SQLContext.scala#L486-L488). Basically, instead of using `CatalystTypeConverters.createToCatalystConverter`, we create a string to scala converters). But, probably it should be in another pr.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -0,0 +1,51 @@
...
+      case IntegerType => datum.toInt
+      case LongType => datum.toLong
+      case FloatType => datum.toFloat
+      case DoubleType => datum.toDouble
+      case BooleanType => datum.toBoolean
+      // TODO(hossein): would be good to support other common timestamp formats
+      case TimestampType => Timestamp.valueOf(datum)
+      // TODO(hossein): would be good to support other common date formats
+      case DateType => Date.valueOf(datum)
+      case StringType => datum
```
yhuai: Seems we may break existing use cases if we do not support `DecimalType`? I think we can create a `java.math.BigDecimal` directly from the string.

rxin:
@yhuai I wouldn't worry too much about performance here. CSV parsing is super slow anyway :)

shivaram:
FWIW I just ran into this bug while trying to apply a schema from SparkR.

falaki:
@yhuai Addressed comments. ptal.

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -0,0 +1,35 @@
...
+  test("Can parse decimal type values") {
+    val stringValues = Seq("10.05", "1,000.01", "158,058,049.001")
+    val decimalValues = Seq(10.05, 1000.01, 158058049.001)
+    val decimalType = new DecimalType(None)
+
+    stringValues.zip(decimalValues).foreach { case (strVal, decimalVal) =>
+      assert(TypeCast.castTo(strVal, decimalType) === new BigDecimal(decimalVal.toString))
+    }
+  }
+}
```
yhuai: new line at here?

yhuai:
just a tiny format thing. LGTM

PR #73, opened 2015-06-04
Manually unescape delimiter character instead of using commons-lang3

falaki:
Instead of depending on commons-lang3 for unescaping java strings, we could manually construct important delimiter characters from the escaped string.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -68,13 +68,12 @@ object TypeCast {
         case '\"' => '\"'
         case '\'' => '\''
         case 'u' if str == """\u0000""" => '\u0000'
-        case '0' => '\0'
-        case _ => throw new RuntimeException(s"Unsupported special character for delimiter: $str")
+        case _ => throw new IllegalArgumentException(s"Unsupported special character for delimiter: $str")
       }
     } else if (str.length == 1) {
       str.charAt(0)
     } else {
-      throw new RuntimeException(s"Delimiter cannot be more than one character: $str")
+      throw new IllegalArgumentException(s"Delimiter cannot be more than one character: $str")
     }
   }
 }
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -57,6 +57,7 @@ object TypeCast {
    * character.
    *
    */
+  @throws[IllegalArgumentException]
   private[csv] def toChar(str: String): Char = {
     if (str.charAt(0) == '\\') {
       str.charAt(1)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -50,4 +50,31 @@ object TypeCast {
...
+       match {
+        case 't' => '\t'
+        case 'r' => '\r'
+        case 'b' => '\b'
+        case 'f' => '\f'
+        case '\"' => '\"'
+        case '\'' => '\''
+        case 'u' if str == """\u0000""" => '\u0000'
+        case '0' => '\0'
+        case _ => throw new RuntimeException(s"Unsupported special character for delimiter: $str")
```
rxin: IllegalArgumentException?

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -50,4 +50,31 @@ object TypeCast {
...
+   */
+  private[csv] def toChar(str: String): Char = {
+    if (str.startsWith("\\")) {
+      str.charAt(1)
+       match {
+        case 't' => '\t'
+        case 'r' => '\r'
+        case 'b' => '\b'
+        case 'f' => '\f'
+        case '\"' => '\"'
```
rxin: isn't this invalid?

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -50,4 +50,31 @@ object TypeCast {
...
+        case '\'' => '\''
+        case 'u' if str == """\u0000""" => '\u0000'
+        case '0' => '\0'
```
rxin: what is \0?

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -50,4 +50,31 @@ object TypeCast {
...
   }
+
+  /**
+   * Helper method that converts string representation of a character to actual character.
+   * It handles some Java escaped strings and throws exception if given string is longer than one
+   * character.
+   *
+   */
+  private[csv] def toChar(str: String): Char = {
+    if (str.startsWith("\\")) {
```
rxin: use char throughout if you are using char. str.charAt(0)

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -32,4 +32,9 @@ class TypeCastSuite extends FunSuite {
       assert(TypeCast.castTo(strVal, decimalType) === new BigDecimal(decimalVal.toString))
     }
   }
+
+  test("Can parse escaped \t") {
+    val escapedStr = """\t"""
```
rxin: if you support other chars you should add test cases for them.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
falaki: Depending on configured quote character, this can be a delimiter character.

rxin: won't user just put in a single character quote? why does it need a \

PR #75, opened 2015-06-05
Support null parsing (empty fields)

smola:


smola:
Note that his used to work in 1.0.3.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -34,6 +34,7 @@ object TypeCast {
    */
   private[csv] def castTo(datum: String, castType: DataType): Any = {
     castType match {
+      case null => null
```
falaki: In what case inferred or passed schema data type is null?

smola: Sorry, I'm not sure anymore. I'm not using this datasource currently.

PR #76, opened 2015-06-08
Configurable null values

petro-rudenko:
There's datasets where each column has it's own marker for missing values. spark-csv assumes only empty string for missing values. To avoid additional data transformation and saving on user's side would be great to specify a set of null markers and replace them to empty string on a library side.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -34,6 +34,7 @@ class CsvParser {
   private var ignoreLeadingWhiteSpace: Boolean = false
   private var ignoreTrailingWhiteSpace: Boolean = false
   private var parserLib: String = ParserLibs.DEFAULT
+  private var nullValues: Seq[String] = Seq("")
 
 
   def withUseHeader(flag: Boolean): CsvParser = {
@@ -81,6 +82,11 @@ class CsvParser {
     this
   }
 
+  def withNullValues(nullValues: Seq[String]): CsvParser = {
+    this.nullValues = nullValues
+    this
+  }
+
   /** Returns a Schema RDD for the given CSV path. */
   @throws[RuntimeException]
   def csvFile(sqlContext: SQLContext, path: String): DataFrame = {
@@ -94,7 +100,8 @@ class CsvParser {
       parserLib,
       ignoreLeadingWhiteSpace,
       ignoreTrailingWhiteSpace,
-      schema)(sqlContext)
+      schema,
+      nullValues)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
   }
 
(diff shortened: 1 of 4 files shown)


saurfang:
+1. This will be very helpful.

falaki:
@petro-rudenko Since this can easily be done with a transformation, I prefer leaving it as it is rather than adding yet another option to spark-csv.

petro-rudenko:
Spark-csv accepts sqlContext and path to files. So transformation is only possible by saving to file, which is not efficient for big files. Also replacement is done on token basis (when csv parser parsed the data). If doing csv parsing on the client side - there would no need to use spark-csv.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -33,21 +33,25 @@ object TypeCast {
...
-      case _: DoubleType => datum.toDouble
-      case _: BooleanType => datum.toBoolean
-      case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
-      // TODO(hossein): would be good to support other common timestamp formats
-      case _: TimestampType => Timestamp.valueOf(datum)
-      // TODO(hossein): would be good to support other common date formats
-      case _: DateType => Date.valueOf(datum)
-      case _: StringType => datum
-      case _ => throw new RuntimeException(s"Unsupported type: ${castType.typeName}")
+    if (datum.isEmpty && castType != StringType) {
```
binarybana: It'd be nice if this was another option. IE: In my application we have decided to standardize on parsing empty string fields as nulls rather than empty strings.

mohitjaggi:
i was working on the same and several other options. see https://github.com/databricks/spark-csv/pull/94/files
i closed that pull request temporarily as i forgot to extract the unit tests from bigdf code.

mohitjaggi:
please look at pull request #113

jaley:
One reason to want this over client-side processing is that user-provided schemata have to initially state the all nullable columns are `StringType` in order to avoid parsing problems. [Amazon ELB logs](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/access-log-collection.html) use a `-` character to represent `null`, including columns that are really `DoubleType` (request time in seconds). You'll get a number format exception if you try to make that a double column straight up.

Seeing as a user-provided schema can tell us whether a column is nullable or not, it might be nice if we could also say what the null values will actually look like in the data.

ragrawal:
+1. I have some data that was generated using R and in this case nulls are encoded as "NA". Currently I am running another job that converts "NA" to "" but it will be nice if there is an option to specify how null values are encoded. All CSV parser I know off have such an option.

HyukjinKwon:
@falaki As #224 is merged, I think we might need to close this?

PR #78, opened 2015-06-11
Increase default max chars per column

petro-rudenko:
For textual data (e.g. sentiment analysis csv - paragraph, score) default univocity limit of max chars per column of 4096 is not enough. It should be configurable, but since this setting is only for univocity parser, could be hardcoded.


falaki:
@petro-rudenko Is this also a problem for the default parser?

petro-rudenko:
No default parser don't have such limit. Only for univocity.

falaki:
Thanks.

PR #82, opened 2015-06-18
handle file encodings other than utf-8

pmackles:
Add support for different file encodings which can be useful when dealing with CSV files from the wild. In our case, we had 8859-1 files containing "þ" which doesn't work with UTF-8. The core of the fix is to add a 'charset' option which defaults to 'UTF-8'. For values other than 'UTF-8', call `SparkContext.hadoopFile()` instead of `SparkContext.textFile()` and translate the `hadoop.io.Text` object to a string based on the specified encoding. Inspiration for this workaround comes from: 

http://stackoverflow.com/questions/15883895/textinputformat-vs-non-utf-8-encoding

This change is only applicable to reading files. I do not think its a good idea to support different charsets when writing files

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -17,14 +17,13 @@ package com.databricks.spark.csv
 
 import java.io.IOException
 
-import org.apache.hadoop.io.{Text, LongWritable}
-import org.apache.hadoop.mapred.TextInputFormat
-
 import scala.collection.JavaConversions._
 import scala.util.control.NonFatal
 
 import org.apache.commons.csv._
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.{Text, LongWritable}
+import org.apache.hadoop.mapred.TextInputFormat
 import org.slf4j.LoggerFactory
 
 import org.apache.spark.rdd.RDD
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -22,15 +22,13 @@ import scala.util.control.NonFatal
 
 import org.apache.commons.csv._
 import org.apache.hadoop.fs.Path
-import org.apache.hadoop.io.{Text, LongWritable}
-import org.apache.hadoop.mapred.TextInputFormat
 import org.slf4j.LoggerFactory
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
 import org.apache.spark.sql.sources.{BaseRelation, InsertableRelation, TableScan}
 import org.apache.spark.sql.types.{StringType, StructField, StructType}
-import com.databricks.spark.csv.util.{ParserLibs, ParseModes, TypeCast}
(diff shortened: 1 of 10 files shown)


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -63,7 +67,14 @@ case class CsvRelation protected[spark] (
 
   // By making this a lazy val we keep the RDD around, amortizing the cost of locating splits.
   def buildScan = {
-    val baseRDD = sqlContext.sparkContext.textFile(location)
+    val baseRDD = if (charset == DefaultCharset) {
```
falaki: Would you please factor out this as a utility function that takes sparkContext and the charset and returns the RDD? Then you can reuse it in line 140 to get first line.

falaki: Also would be good if we could a unit test for the utility function.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -102,6 +102,8 @@ class DefaultSource
       throw new Exception("Ignore white space flag can be true or false")
     }
 
+    val charset = parameters.getOrElse("charset", DefaultCharset)
```
falaki: I think it is a good idea to validate the charset. Either here or in the utility function that uses it to read data into an RDD.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -22,6 +22,8 @@ import org.apache.spark.sql.{DataFrame, SQLContext}
 
 package object csv {
 
+  val DefaultCharset = "UTF-8"
```
falaki: Can we use ```java.nio.charset.StandardCharsets``` instead of the string literals?

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -15,7 +15,7 @@
  */
 package com.databricks.spark.csv
 
-import java.io.File
+import java.io.{File}
```
falaki: Do you need additional classes from ```java.io```?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -17,6 +17,9 @@ package com.databricks.spark.csv
 
 import java.io.IOException
 
+import org.apache.hadoop.io.{Text, LongWritable}
+import org.apache.hadoop.mapred.TextInputFormat
```
falaki: please order imports according to Spark style guide: https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide#SparkCodeStyleGuide-Imports

pmackles:
@falaki - thanks for the feedback. I think the last two commits address your concerns.

falaki:
@pmackles Thanks for addressing the comments. Would you please take a look at the scala style failure.

pmackles:
@falaki - fixed the scalastyle failure. I did have to remove the references to java.nio.charset.StandardCharsets to get it to build with JDK1.6.

falaki:
Thanks. This looks good. Merging it.

PR #93, opened 2015-07-07
Add schema inference

### LINKED ISSUE #60: Data types not inferred
Spark SQL has the ability to infer column types from, say, JSON files. It would be nice if spark-csv did too for CSV files. Currently, in Python at least, all types are strings (with or without header="true").

dtpeacock:
This resolves #60 , the schema is inferred in a similar way to how the JSON schema is inferred.

The public api has not changed other than adding the inferSchema option, e.g.:
    
    csvFile(carsFile, inferSchema = true)

    OPTIONS (path "path", inferSchema "true")

I have pulled the tokenizing code into a common method as the schema has to be inferred on the tokenized data.

Currently this uses the entire dataset to discover the schema, I could also add on a sampling `samplingRatio` option as in the JSON schema inference?

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -35,6 +35,7 @@ class CsvParser {
   private var ignoreTrailingWhiteSpace: Boolean = false
   private var parserLib: String = ParserLibs.DEFAULT
   private var charset: String = TextFile.DEFAULT_CHARSET.name()
+  private var inferSchema: Boolean = false
 
   def withUseHeader(flag: Boolean): CsvParser = {
     this.useHeader = flag
@@ -86,6 +87,11 @@ class CsvParser {
     this
   }
 
+  def withInferSchema(inferSchema: Boolean) = {
+    this.inferSchema = inferSchema
+    this
+  }
+
   /** Returns a Schema RDD for the given CSV path. */
   @throws[RuntimeException]
   def csvFile(sqlContext: SQLContext, path: String): DataFrame = {
@@ -100,7 +106,8 @@ class CsvParser {
       ignoreLeadingWhiteSpace,
       ignoreTrailingWhiteSpace,
       schema,
-      charset)(sqlContext)
+      charset,
+      inferSchema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
   }
 
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -90,12 +90,10 @@ case class CsvRelation protected[spark] (
         parseCSV(csvIter, csvFormat)
       }
     }
-
   }
 
   // By making this a lazy val we keep the RDD around, amortizing the cost of locating splits.
   def buildScan = {
-
     val schemaFields = schema.fields
     tokenRdd(schemaFields.map(_.name)).flatMap{ tokens =>
 
@@ -124,15 +122,12 @@ case class CsvRelation protected[spark] (
         }
       }
     }
-
   }
 
   private def inferSchema(): StructType = {
-
     if (this.userSchema != null) {
       userSchema
     } else {
-
       val firstRow = if(ParserLibs.isUnivocityLib(parserLib)) {
         val escapeVal = if(escape == null) '\\' else escape.charValue()
         new LineCsvReader(fieldSep = delimiter, quote = quote, escape = escapeVal)
@@ -145,13 +140,11 @@ case class CsvRelation protected[spark] (
           .withSkipHeaderRecord(false)
         CSVParser.parse(firstLine, csvFormat).getRecords.head.toArray
       }
-
       val header = if (useHeader) {
         firstRow
       } else {
         firstRow.zipWithIndex.map { case (value, index) => s"C$index"}
       }
-
       if (this.inferCsvSchema) {
         InferSchema(tokenRdd(header), header)
       } else{
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -27,8 +27,8 @@ import org.slf4j.LoggerFactory
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
 import org.apache.spark.sql.sources.{BaseRelation, InsertableRelation, TableScan}
-import org.apache.spark.sql.types.{StringType, StructField, StructType}
-import com.databricks.spark.csv.util.{ParserLibs, ParseModes, TextFile, TypeCast}
+import org.apache.spark.sql.types._
+import com.databricks.spark.csv.util._
 import com.databricks.spark.sql.readers._
 
 case class CsvRelation protected[spark] (
@@ -42,7 +42,8 @@ case class CsvRelation protected[spark] (
     ignoreLeadingWhiteSpace: Boolean,
     ignoreTrailingWhiteSpace: Boolean,
     userSchema: StructType = null,
-    charset: String = TextFile.DEFAULT_CHARSET.name())(@transient val sqlContext: SQLContext)
+    charset: String = TextFile.DEFAULT_CHARSET.name(),
+    inferCsvSchema: Boolean)(@transient val sqlContext: SQLContext)
   extends BaseRelation with TableScan with InsertableRelation {
 
   private val logger = LoggerFactory.getLogger(CsvRelation.getClass)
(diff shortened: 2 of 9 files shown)


src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,73 @@
...
+ */
+package com.databricks.spark.csv.util
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.analysis.HiveTypeCoercion
+import org.apache.spark.sql.types._
+
+import scala.util.control.Exception._
+
+private[csv] object InferSchema {
```
mohitjaggi: thanks. schema inference would be awesome...i wonder what the performance would be like, as exceptions can be expensive. on the other hand you can see in bigdf code https://github.com/AyasdiOpenSource/bigdf/blob/master/src/main/scala/com/ayasdi/bigdf/SchemaUtils.scala this is what i did :-) an option to use sampling instead of using the whole dataset would be nice. i am happy to extract that code from bigdf and contribute

possible optimizations:
1) let inferField know what we have so far, so it can avoid doing unnecessary work. e.g. if we got a double, no need to try parsing as int or long. if we got a string, no need to parse as double.
2) use mappartitions to do an "imperative-style" loop per partition and then merge the partitions (which can be done on the driver as there will be a single schema per partition). this will reduce memory allocation and GC overhead.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,73 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.databricks.spark.csv.util
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.analysis.HiveTypeCoercion
```
falaki: This is not a public API and using it will cause problems if Spark SQL changes. Would you include a simple version of it in your patch? You can copy/paste that code as well. CC @yhuai for comment.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -88,15 +87,52 @@ case class CsvRelation protected[spark] (
         } else {
           iter
         }
-        parseCSV(csvIter, csvFormat, schema.fields)
+        parseCSV(csvIter, csvFormat)
+      }
+    }
+
```
falaki: extra line.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -88,15 +87,52 @@ case class CsvRelation protected[spark] (
...
+  }
+
+  // By making this a lazy val we keep the RDD around, amortizing the cost of locating splits.
+  def buildScan = {
+
```
falaki: extra line.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -88,15 +87,52 @@ case class CsvRelation protected[spark] (
...
+          case aiob: ArrayIndexOutOfBoundsException if permissive =>
+            (index until schemaFields.length).foreach(ind => rowArray(ind) = null)
+            Some(Row.fromSeq(rowArray))
+          case NonFatal(e) if !failFast =>
+            logger.error(s"Exception while parsing line: $tokens. ", e)
+            None
+        }
       }
     }
+
```
falaki: remove extra line.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,73 @@
...
+    } else if ((allCatch opt field.toLong).isDefined) {
+      LongType
+    } else if ((allCatch opt field.toDouble).isDefined) {
+      DoubleType
+    } else {
+      StringType
+    }
+
+  }
+
```
falaki: extra line.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
falaki: I like both suggestions. I suggest doing both, although most of the benefit will come from number 2.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,73 @@
...
+private[csv] object InferSchema {
+
+  /**
+   * Similar to the JSON schema inference. {@link org.apache.spark.sql.json.InferSchema}
+   *     1. Infer type of each row
+   *     2. Merge row types to find common type
+   *     3. Replace any null types with string type
+   */
+  def apply(
+             tokenRdd: RDD[Array[String]],
```
falaki: Is more readable in a single line.

falaki:
Thanks @dtpeacock for submitting this. We really need this feature. I am somewhat concerned with using internal Spark API. I also left a few minor style related comments. @mohitjaggi suggestions for optimization are also good to have. I would leave the extra option for sampling for a separate PR.

dtpeacock:
Thanks for the comments - I have fixed the style issues and pulled the internal spark method across.

Mapping over each partition and then merging each partition's results is definitely a much better way of doing it - I've replaced the original method with an aggregate which does exactly that. 

As suggested, I've also passed through to inferField the type so far, and skipped unneeded cases, although there might be a better way to implement it.

JoshRosen:
I'm experimenting with integrating a code coverage tool into the spark-csv build.  In order to trigger your PR to be re-build with that tool, I'm going to temporarily close it, then will re-open it in order to convince Travis to rebuild the new merge commit.

JoshRosen:
Cool, it worked!  The current build status is a little overzealous relative to our current coverage, so I'm going to disable it for now so as not to get false-positive build failures.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,127 @@
...
+   * Copied from internal Spark api
+   * {@link org.apache.spark.sql.catalyst.analysis.HiveTypeCoercion}
+   */
+  val findTightestCommonType: (DataType, DataType) => Option[DataType] = {
+    case (t1, t2) if t1 == t2 => Some(t1)
+    case (NullType, t1) => Some(t1)
+    case (t1, NullType) => Some(t1)
+
+    // Promote numeric types to the highest of the two and all numeric types to unlimited decimal
+    case (t1, t2) if Seq(t1, t2).forall(numericPrecedence.contains) =>
```
JoshRosen: Codecov claims that this case and all of the following cases are aren't covered in tests.

src/main/scala/com/databricks/spark/csv/CsvParser.scala
```diff
@@ -86,6 +87,11 @@ class CsvParser {
     this
   }
 
+  def withInferSchema(inferSchema: Boolean) = {
```
JoshRosen: This isn't called anywhere (Codecov caught this).

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -88,7 +87,39 @@ case class CsvRelation protected[spark] (
...
+          while (index < schemaFields.length) {
+            rowArray(index) = TypeCast.castTo(tokens(index), schemaFields(index).dataType)
+            index = index + 1
+          }
+          Some(Row.fromSeq(rowArray))
+        } catch {
+          case aiob: ArrayIndexOutOfBoundsException if permissive =>
+            (index until schemaFields.length).foreach(ind => rowArray(ind) = null)
+            Some(Row.fromSeq(rowArray))
+          case NonFatal(e) if !failFast =>
```
JoshRosen: This error-handling case is never hit in unit tests.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    val stuctFields = header.zip(rootTypes).map { case (thisHeader, rootType) =>
+      StructField(thisHeader, rootType, nullable = true)
+    }
+
+    StructType(stuctFields)
+  }
+
+  private[csv] def inferRowType(rowSoFar: Array[DataType], next: Array[String]) = {
+    rowSoFar.zipAll(next, NullType, null).map{case (tpe, field) => inferField(tpe, field)}
```
mohitjaggi: the inferField enhancement looks ok...however i am not sure if inferRowType() will avoid the GC issue as zipAll and map will create temp objects. rdd.aggregate() allows you to have a "mutable" left param...so you could do something like this:

var i = 0
while(i < rowSoFar.length) {
 rowSoFar(i) = inferField(rowSoFar(i), field) // not sure if findTightestCommonType needs  be called here?
}

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -0,0 +1,127 @@
...
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.types._
+
+import scala.util.control.Exception._
+
+private[csv] object InferSchema {
+
+  /**
+   * Similar to the JSON schema inference. {@link org.apache.spark.sql.json.InferSchema}
```
JoshRosen: Will `@link` render correctly in Scaladoc?

dtpeacock:
Codecov is pretty nice!

I've removed the unused code, and increased the test coverage.

@mohitjaggi  - does that look better? That should avoid the gc issue now.

mohitjaggi:
@dtpeacock optimization itself lgtm. from a code factoring perspective: is there a way to have the "merging" logic in a single place yet have the optimization of avoiding unnecessary exception catching?

falaki:
Thanks @dtpeacock merging this.

PR #100, opened 2015-07-14
Update the version to 1.1.0 and add artifactId for Scala 2.10 in README.

yhuai:
Seems our README still uses 1.0.3, which is not compatible with Spark 1.4. It will be better to update the version to 1.1.0. 

btw, 1.1.0 works with Spark 1.3, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,25 @@ This library requires Spark 1.3+
 ## Linking
 You can link against this library in your program at the following coordiates:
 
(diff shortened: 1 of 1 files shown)


falaki:
Thanks @yhuai. I think we released spark-csv 1.1 with Spark 1.4. Did we add new features to data sources API in 1.4? In that case spark-csv 1.1 would not work with spark 1.3

Issue #86, opened 2015-06-22
Missing Values in CSV is taking as blank String not null


abhishek-ch:
I am trying to load the csv file using pyspark and found blank values in CSV is always taken as blank String in Dataframe created even in PERMISSIVE mode but the doc says 

>PERMISSIVE: tries to parse all lines: nulls are inserted for missing tokens and extra tokens are ignored.

I am using following script -
```
df_test= sqlContext.read.format("com.databricks.spark.csv").options(delimiter=",").options(header="true").load("/Volumes/test.csv")
```
![image](https://cloud.githubusercontent.com/assets/7579608/8291106/3ce9a6e6-191f-11e5-8ff7-22e8d0fe4159.png)

thats why now fillna is not working .

falaki:
What does the raw data look like? Just want to make sure there is no space or other character in place of that token.

abhishek-ch:
its a blank cell with no space & character in a csv cell.

There is a piece of code in CSVReader
settings.setNullValue("")

I tried modifying it to null and it didn't work , it was bit hard to understand the code as not aware of scala.

1stdayonthejob:
I'm using the cars.csv from the readme and am also getting blank results.

~~~python
df = sqlContext.read.format('com.databricks.spark.csv').options(header='true').load('cars.csv')
df.fillna('SHOULD BE NULL').show()
~~~

~~~
+----+-----+-----+--------------------+--------------+
|year| make|model|             comment|         blank|
+----+-----+-----+--------------------+--------------+
|2012|Tesla|    S|          No comment|              |
|1997| Ford| E350|Go get one now th...|SHOULD BE NULL|
|2015|Chevy| Volt|      SHOULD BE NULL|SHOULD BE NULL|
+----+-----+-----+--------------------+--------------+
~~~

I would expect the 'blank' column in the Tesla line to be null.

Also if I add commas in the cars.csv file to have it separate all columns

~~~
year,make,model,comment,blank
"2012","Tesla","S","No comment",

1997,Ford,E350,"Go get one now they are going fast",
2015,Chevy,Volt,,
~~~

then all the fields are emty string rather than nulls

~~~
+----+-----+-----+--------------------+-----+
|year| make|model|             comment|blank|
+----+-----+-----+--------------------+-----+
|2012|Tesla|    S|          No comment|     |
|1997| Ford| E350|Go get one now th...|     |
|2015|Chevy| Volt|                    |     |
[...]

PR #102, opened 2015-07-16
Support for nullable schema types

### LINKED ISSUE #85: No failover for empty fields which can not be parsed by a given schema
A CSV line like:
* a;b;c;d;;e;f => an empty field between d & e
* a;b;c;d;e; => an empty field in the end

When an empty field is encountered, it is considered as an empty String "".
If a schema given for the field is, for example, an `IntegerType`, DataFrameRead will throw a `NumberFormatException` which makes the line excluded.

It would be better:
When a field is considered as an empty String, just return `null` for the basic data types (Integer, Long, etc).

### LINKED ISSUE #97: spark-csv fails to load csv with empty fields
Environment: spark 1.4 with com.databricks:spark-csv_2.10:1.1.0

spark-csv fails to load csv files with empty fields, even if I declare nullable=True in my schema definition:

```python
from pyspark.sql import SQLContext
from pyspark.sql.types import StructType, StructField, IntegerType, StringType

sqlContext = SQLContext(sc)
schema = StructType([
    StructField('id', IntegerType(), True),
    StructField('name', StringType(), True),
    StructField('value', IntegerType(), True)
])
df = sqlContext.read.format('com.databricks.spark.csv').schema(schema).options(header='true', mode="FAILFAST").load('hdfs://XXX/test.csv')
df.collect()
```

csv file (first entry has no value set)

```csv
id,name,value
1,foo,
2,bar,0
```

Exception
```
Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 1.0 failed 4 times, most recent failure: Lost task 1.3 in stage 1.0 (TID 7, 10.2.160.14): java.lang.NumberFormatException: For input string: ""
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
```

Expected behaviour is imho that there are inserted null values instead of raising an exception. Eg add something like

```scala
if (schemaFields(index).nullable && tokens(index) == ""){
                 rowArray(index) = null
}
```

to 

https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/CsvRelation.scala#L156

and

https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/CsvRelation.scala#L198

dtpeacock:
This closes #85 and closes #97 .

This seems to be most sensible behavior - if the schema specifies a nullable type, it should return null on an empty field.

To maintain backwards compatibility, I have left it so that in the case of a StringType, "" is passed back rather than null  (i.e. fieldname='' does not have to be changed to fieldname=null).

Its worth confirming what the desired behavior should be here as in #86  - should "" be treated as null or an empty string, as here would be the place to change it.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -108,14 +108,8 @@ case class CsvRelation protected[spark] (
         try {
           index = 0
           while (index < schemaFields.length) {
-            rowArray(index) = if (schemaFields(index).nullable && tokens(index) == ""){
-              schemaFields(index).dataType match {
-                case StringType => ""
-                case _ => null
-              }
-            } else {
-              TypeCast.castTo(tokens(index), schemaFields(index).dataType)
-            }
+            val field = schemaFields(index)
+            rowArray(index) = TypeCast.castTo(tokens(index), field.dataType, field.nullable)
             index = index + 1
           }
           Some(Row.fromSeq(rowArray))
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -108,7 +108,14 @@ case class CsvRelation protected[spark] (
         try {
           index = 0
           while (index < schemaFields.length) {
-            rowArray(index) = TypeCast.castTo(tokens(index), schemaFields(index).dataType)
+            rowArray(index) = if (schemaFields(index).nullable && tokens(index) == ""){
+              schemaFields(index).dataType match {
+                case StringType => ""
+                case _ => null
+              }
+            } else {
+              TypeCast.castTo(tokens(index), schemaFields(index).dataType)
+            }
             index = index + 1
           }
           Some(Row.fromSeq(rowArray))
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -29,25 +29,33 @@ object TypeCast {
    * Casts given string datum to specified type.
    * Currently we do not support complex types (ArrayType, MapType, StructType).
    *
+   * For string types, this is simply the datum. For other types.
+   * For other nullable types, this is null if the string datum is empty.
+   *
    * @param datum string value
    * @param castType SparkSQL type
    */
-  private[csv] def castTo(datum: String, castType: DataType): Any = {
-    castType match {
-      case _: ByteType => datum.toByte
-      case _: ShortType => datum.toShort
-      case _: IntegerType => datum.toInt
-      case _: LongType => datum.toLong
-      case _: FloatType => datum.toFloat
-      case _: DoubleType => datum.toDouble
-      case _: BooleanType => datum.toBoolean
-      case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
-      // TODO(hossein): would be good to support other common timestamp formats
-      case _: TimestampType => Timestamp.valueOf(datum)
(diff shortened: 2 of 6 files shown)


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -108,7 +108,14 @@ case class CsvRelation protected[spark] (
         try {
           index = 0
           while (index < schemaFields.length) {
-            rowArray(index) = TypeCast.castTo(tokens(index), schemaFields(index).dataType)
+            rowArray(index) = if (schemaFields(index).nullable && tokens(index) == ""){
```
falaki: I don't think we need to check for nullability for StringType. Basically I think the following logic would be simple an intuitive:
* if StringType just return the token (don't need to check for nullability)
* if any other time, if token is "", return null, else cast.

I suggest moving this logic (along with some comment that explains it) to TypeCast.castTo (maybe as a simple private method). This way you can add a few simple unit tests for it.

falaki:
Thanks @dtpeacock for submitting this. I left one comment, please address that and I will merge it. I am planning to cut a release after this.

dtpeacock:
Thanks, I've addressed the comment and added a couple more tests!

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -29,25 +29,33 @@ object TypeCast {
...
-      case _: BooleanType => datum.toBoolean
-      case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
-      // TODO(hossein): would be good to support other common timestamp formats
-      case _: TimestampType => Timestamp.valueOf(datum)
-      // TODO(hossein): would be good to support other common date formats
-      case _: DateType => Date.valueOf(datum)
-      case _: StringType => datum
-      case _ => throw new RuntimeException(s"Unsupported type: ${castType.typeName}")
+  private[csv] def castTo(datum: String, castType: DataType, nullable: Boolean = true): Any = {
+    if (castType.isInstanceOf[StringType]){
```
falaki: I think this would be simpler:
```scala
if (datum == "" && nullable && !castType.isInstanceOf[StringType]) {
  null
} else {
  castType match {
      case _: ByteType => datum.toByte
      case _: ShortType => datum.toShort
      case _: IntegerType => datum.toInt
      case _: LongType => datum.toLong
      case _: FloatType => datum.toFloat
      case _: DoubleType => datum.toDouble
      case _: BooleanType => datum.toBoolean
      case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
      // TODO(hossein): would be good to support other common timestamp formats
      case _: TimestampType => Timestamp.valueOf(datum)
      // TODO(hossein): would be good to support other common date formats
      case _: DateType => Date.valueOf(datum)
      case _: StringType => datum
      case _ => throw new RuntimeException(s"Unsupported type: ${castType.typeName}")
  }
}
```

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -56,4 +57,33 @@ class TypeCastSuite extends FunSuite {
...
+  }
+
+  test("Throws exception for empty string with non null type"){
+    val exception = intercept[NumberFormatException]{
+      TypeCast.castTo("", IntegerType, nullable = false)
+    }
+    assert(exception.getMessage.contains("For input string: \"\""))
+  }
+
+  test("Types are cast correctly"){
```
falaki: Thanks for adding these!

dtpeacock:
Agreed that logic is a bit simpler

PR #105, opened 2015-07-22
Support comment characters

jaley:
Many automatically generated TSV files, such as logs from [Amazon CloudFront](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html) use a `#` character for comments (used to put version information or sometimes even headers) at the top of the file. The CSV parser actually supports this already, so this change just connects it up to the parameters.

The requirement to handle comments at the start of the file is actually a bit problematic, because CsvReation actually peeks at the first line to infer the schema if it's not explicitly provided. The implementation I went with was to pick some fixed maximum number of lines we will peek ahead searching for a line that isn't a comment before failing, if commentMarker is provided. Not entirely happy with that as it feels a bit nasty, but not sure what else could be done - suggestions welcome!


jaley:
Travis failure is apparently because sbt download failed? Is that just a random network failure? I don't think I've changed anything build related! :)

falaki:
@jaley Thanks for this patch. Would you please add the unit test to ```CsvFastSuite.scala``` as well and make sure this new feature works with both parsers. Also would you update ```README.md``` to include the new option.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -46,6 +47,11 @@ case class CsvRelation protected[spark] (
     inferCsvSchema: Boolean)(@transient val sqlContext: SQLContext)
   extends BaseRelation with TableScan with InsertableRelation {
 
+  /**
+   * Limit the number of lines we'll search for a header row that isn't comment-prefixed.
+   */
+  private val MAX_COMMENT_LINES_IN_HEADER = 100
```
falaki: I feel 100 lines may be excessive. How many comment lines do you usually see in your use cases? Alternatively, we can set a smaller default and expose it as an option to be increased.

jaley:
@falaki Thanks for the review feedback, I think I've actioned all your comments in the last commit. One thing to note though...

The default behaviour for comments was actually already different when using Univocity vs. commons - though I didn't realize initially. Univocity actually supports `#` comments by default, whereas commons doesn't support comments without you explicitly setting a comment character. Seeing as `spark-csv` didn't change the setting on either parser initially, behaviour in current releases will be that univocity supports comments by default and commons doesn't.

So, if we were to make the behaviour consistent now for both parsers, we'd technically be breaking compatibility and may cause people's code to break unless they start using the new comment parameter, whereas previously they didn't need to. For that reason, I connected up the parameter to univocity in a way that preserves the existing behaviour, and just documented the differences in the README.

The drawback is that this means there's no sensible way to explicitly disable comment characters on univocity, as `null` means "use the default" in the current implementation. Your call on what to do here - we either clean up to make the behaviour consistent and break compatibility, or go with this limitation I guess?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -159,8 +166,17 @@ case class CsvRelation protected[spark] (
...
-    TextFile.withCharset(sqlContext.sparkContext, location, charset).first()
-  }
+    val csv = TextFile.withCharset(sqlContext.sparkContext, location, charset)
+    if(commentMarker == null) {
+      csv.first()
+    } else {
+      csv.take(MAX_COMMENT_LINES_IN_HEADER)
+        .find(! _.startsWith(commentMarker.toString)) match {
+        case Some(line) => line
+        case None => throw new RuntimeException(s"No uncommented header line in first $MAX_COMMENT_LINES_IN_HEADER lines")
```
falaki: This line exceeds 100 characters (it fails the style check).

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -32,6 +32,7 @@ package object csv {
                 delimiter: Char = ',',
                 quote: Char = '"',
                 escape: Character = null,
+                comment: Character = null,
```
falaki: We are getting a style error here. That is because we are exceeding 11 parameters. I am fixing that in a separate PR.

README.md
```diff
@@ -46,6 +46,9 @@ When reading files the API accepts several options:
   * `FAILFAST`: aborts with a RuntimeException if encounters any malformed line
 * `charset`: defaults to 'UTF-8' but can be set to other valid charset names
 * `inferSchema`: automatically infers column types. It requires one extra pass over the data and is false by default
+* `comment`: skip lines beginning with this character. Note the default behaviours are different for each parser:
```
falaki: I discussed it with @rxin and @JoshRosen. In this case, it seems keeping a uniform behavior is a better choice. So please update the code to have ```#``` as comment character for both parsers and update the doc.

jaley:
@falaki  Addressed review comments, but please hold off merging. I've added a test case to check that nulling out the comment character actually disables comments, and it doesn't for the fast parser yet. I'm working on a fix and will update tomorrow. Hopefully that'll also increase coverage :-)

jaley:
@falaki OK, I fixed that problem now. It's now possible to disable comments by setting the parameter to null. The default is `#` for both parsers.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -67,6 +67,15 @@ class DefaultSource
       throw new Exception("Escape character cannot be more than one character.")
     }
 
+    val comment = parameters.getOrElse("comment", "#")
+    val commentChar: Character = if (comment == null) {
```
JoshRosen: Since you used `getOrElse` on the previous line, isn't it now impossible for `comment` to be `null`?

src/main/scala/com/databricks/spark/csv/package.scala
JoshRosen: We can also disable that style checking rule if you think it's not appropriate.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -159,8 +167,19 @@ case class CsvRelation protected[spark] (
...
    */
   private lazy val firstLine = {
-    TextFile.withCharset(sqlContext.sparkContext, location, charset).first()
-  }
+    val csv = TextFile.withCharset(sqlContext.sparkContext, location, charset)
+    if(comment == null) {
+      csv.first()
+    } else {
+      csv.take(MAX_COMMENT_LINES_IN_HEADER)
+        .find(! _.startsWith(comment.toString)) match {
```
JoshRosen: You could use `getOrElse` instead of this match.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -159,8 +167,19 @@ case class CsvRelation protected[spark] (
    * Returns the first line of the first non-empty file in path
    */
   private lazy val firstLine = {
-    TextFile.withCharset(sqlContext.sparkContext, location, charset).first()
-  }
+    val csv = TextFile.withCharset(sqlContext.sparkContext, location, charset)
+    if(comment == null) {
```
JoshRosen: Space after `if` keyword.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
jaley: No, I don't think. We'll default to `#` only if the user didn't set a value for `comment` at all. If they've explicitly set it to `null`, that will still return `null`:

```scala
scala> Map("comment" -> null).getOrElse("comment", "#")
res1: String = null
```

jaley:
@JoshRosen Thanks for comments - updated.

falaki:
@jaley Thanks for this. I am merging it.

PR #113, opened 2015-07-30
parsing options and serializing arrays

mohitjaggi:
several parsing options are added. they are organized in classes because there are many of them. a "text" based API to configure options is provided.

another feature is the ability to serialize a column of arrays. the array is "unnested" and column names to use are supplied by user. this is useful for writing out csv after doing transforms on the data that "expand" the number of columns e.g. one hot encode a category. this can be improved later. e.g. sparse vector from mllib can replace the array.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,20 +12,9 @@ libraryDependencies += "org.apache.commons" % "commons-csv" % "1.1"
 
 libraryDependencies += "com.univocity" % "univocity-parsers" % "1.5.1"
 
-libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.5" % "provided"
+libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.5" % Provided
 
-libraryDependencies ++= Seq(
-  ("org.apache.spark" %% "spark-sql" % "1.4.0").
-//    exclude("org.mortbay.jetty", "servlet-api").
-    exclude("commons-beanutils", "commons-beanutils-core").
-    exclude("commons-collections", "commons-collections").
-    exclude("commons-logging", "commons-logging").
-    exclude("org.slf4j", "slf4j-api").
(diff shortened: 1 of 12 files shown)


build.sbt
```diff
@@ -1,20 +1,20 @@
 name := "spark-csv"
 
-version := "1.1.0"
+version := "1.1.0-SNAPSHOT"
```
falaki: Why do you need this?

build.sbt
```diff
@@ -1,20 +1,20 @@
...
 
 organization := "com.databricks"
 
-scalaVersion := "2.11.6"
+scalaVersion := "2.10.4"
```
falaki: Why do you need scala 2.10? You know we publish jars for both version?

build.sbt
```diff
@@ -24,45 +24,6 @@ resolvers ++= Seq(
 
 publishMavenStyle := true
 
-spAppendScalaVersion := true
```
falaki: Why removing this?

mohitjaggi: please ignore build.sbt

falaki:
Please update it to rebase against master for tests to run. Also please revert changes to ```build.sbt```.

mohitjaggi:
@falaki should be ready to merge now.

falaki:
@mohitjaggi This is fairly large. I am about to publish a release with schema inference and all the recent improvements, and then I will review this.

falaki:
@mohitjaggi this is packing too much into a single PR. Would you please split it. Please first submit one for number parsing improvements and another for arrays. On arrays within CSV it would be good to post an issue and gather some feedback from community first.

mohitjaggi:
will do

mohitjaggi:
see #124

HyukjinKwon:
Wouldn't this make sense to close this if #124 is subset from this and you are willing to make some more?

PR #124, opened 2015-08-13
options changes only

mohitjaggi:


src/main/scala/com/databricks/spark/csv/CsvParser.scala
```diff
@@ -16,24 +16,24 @@
...
 /**
  * A collection of static functions for working with CSV files in Spark SQL
  */
 class CsvParser {
-
   private var useHeader: Boolean = false
-  private var delimiter: Character = ','
-  private var quote: Character = '"'
-  private var escape: Character = null
+  private var csvParsingOpts: CSVParsingOpts = CSVParsingOpts()
```
falaki: How come comment, schema, and charset are not included in CSVParsingOpts? I don't clearly see a benefit for this level of indirection yet.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -113,36 +114,52 @@ class DefaultSource
...
-      parseMode,
-      parserLib,
-      ignoreLeadingWhiteSpaceFlag,
-      ignoreTrailingWhiteSpaceFlag,
-      schema,
-      charset,
-      inferSchemaFlag)(sqlContext)
+      useHeader = headerFlag,
+      csvParsingOpts = csvParsingOpts,
+      lineExceptionPolicy = lineParsingOpts,
```
falaki: What does this additional parameter accomplish that parsing mode does not?

src/main/scala/com/databricks/spark/csv/CsvParser.scala
mohitjaggi: as the parser improves a ton of options are being added. it makes sense to organize them in containers for ease of use. imho, schema is a separate thing. i intended csvparsingopts to contain options that are pass to the csv parsing library. i can change the name if there is a better one you can think of

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
mohitjaggi: should be the same. renamed it to indicate that this is a "per line" exception handling policy.

HyukjinKwon:
@mohitjaggi I think some of them are fixed already and the changes are too big without some descriptions to follow. Wouldn't this be ensible if you close this and maybe submit a PR with smaller one? Or rebase this if you are working on this?

PR #125, opened 2015-08-13
Adding csvRdd method which allows parsing csvs from a preexisting rdd.

### LINKED ISSUE #62: missing csvRDD function
[SQLContext](https://spark.apache.org/docs/1.3.1/api/scala/index.html#org.apache.spark.sql.SQLContext) has a `jsonRDD(json: RDD[String]): DataFrame`

### LINKED ISSUE #91: Read csv from existing RDD[String] rather than from filesystem path?
Is there a simple way to create a Dataframe from an RDD[String], where each element of the RDD is a row from a CSV file?  We have some malformed CSVs that we're trying to work with, and being able to do something like
```scala
def fix(line: String): Option[String] = ...

val df = (sc.textFile(...) flatMap fix).readCsv
```
(or something similar), without the intermediate step of creating new, fixed CSVs, would be really helpful.

dtpeacock:
Closes #91 and closes #62 . This allows an existing string RDD to be loaded as a csv with:
 `new CsvParser().csvRdd(rdd)`

I've pulled out the path from the CsvRelation so that it is instead passed through as an RDD. This is passed through as a function so that it's is recreated on every operation, which is also how the json relation works.

I've had to make CsvParser and CsvContext serializable - this is because the anonymous functions are declared there. The anonymous function could be moved into the CsvRelation constructor?

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -115,7 +115,6 @@ class CsvParser extends Serializable {
       ignoreLeadingWhiteSpace,
       ignoreTrailingWhiteSpace,
       schema,
-      charset,
       inferSchema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
   }
@@ -135,7 +134,6 @@ class CsvParser extends Serializable {
       ignoreLeadingWhiteSpace,
       ignoreTrailingWhiteSpace,
       schema,
-      charset,
       inferSchema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
 
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -16,6 +16,7 @@
 package com.databricks.spark.csv
 
 
+import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{DataFrame, SQLContext}
 import org.apache.spark.sql.types.StructType
 import com.databricks.spark.csv.util.{ParserLibs, ParseModes, TextFile}
@@ -23,7 +24,7 @@ import com.databricks.spark.csv.util.{ParserLibs, ParseModes, TextFile}
 /**
  * A collection of static functions for working with CSV files in Spark SQL
  */
-class CsvParser {
+class CsvParser extends Serializable {
 
   private var useHeader: Boolean = false
   private var delimiter: Character = ','
@@ -102,7 +103,8 @@ class CsvParser {
   @throws[RuntimeException]
   def csvFile(sqlContext: SQLContext, path: String): DataFrame = {
     val relation: CsvRelation = CsvRelation(
-      path,
+      () => TextFile.withCharset(sqlContext.sparkContext, path, charset),
+      Some(path),
       useHeader,
       delimiter,
       quote,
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -32,7 +32,8 @@ import com.databricks.spark.csv.util._
 import com.databricks.spark.sql.readers._
 
 case class CsvRelation protected[spark] (
-    location: String,
+    baseRDD: () => RDD[String],
```
falaki: With this change, we don't need to pass charset to CsvRelation any more. You can remove it from list of arguments.

src/main/scala/com/databricks/spark/csv/CsvParser.scala
```diff
@@ -118,5 +120,24 @@ class CsvParser {
...
+      delimiter,
+      quote,
+      escape,
+      comment,
+      parseMode,
+      parserLib,
+      ignoreLeadingWhiteSpace,
+      ignoreTrailingWhiteSpace,
+      schema,
+      charset,
```
falaki: Can be removed after you update ```CsvRelation``` definition. Please also remove form other places where we pass ```charset``` when constructing ```CsvRelation```.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -17,8 +17,9 @@ package com.databricks.spark
 
 import org.apache.commons.csv.CSVFormat
 import org.apache.hadoop.io.compress.CompressionCodec
+import org.apache.spark.rdd.RDD
 
-import org.apache.spark.sql.{DataFrame, SQLContext}
+import org.apache.spark.sql.{DataFrameReader, DataFrame, SQLContext}
```
falaki: Do you need ```DataFrameReader```?

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -171,4 +174,5 @@ package object csv {
       }
     }
   }
+
```
falaki: extra new line.

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -441,4 +441,39 @@ class CsvSuite extends FunSuite {
...
+    df.registerTempTable("csvRdd")
+    sampleDf.registerTempTable("sampleDf")
+
+    val exception = intercept[java.io.IOException] {
+      sql("INSERT OVERWRITE TABLE csvRdd select * from sampleDf")
+    }
+    assert(exception.getMessage.contains("Cannot INSERT into table with no path defined"))
+  }
+
+  test("DSL tsv test") {
```
falaki: Thanks for adding this!

falaki:
@dtpeacock this looks pretty good. Left a few minor comments. And thanks for increasing test coverage.

dtpeacock:
Cheers, have addressed those issues!

summerela:
Hello! I'm so glad you worked on this amazing CSV parser; thank you! 

I work in the field of bioinformatics and often deal with large human genome variant files that begin with upwards of 200 commented fields, starting with '#'. 

The following value is hard-coded in your parser: MAX_COMMENT_LINES_IN_HEADER= 10 and renders this CSV parser unusable for a lot of biological data. Is there a way for me to set this variable when I read in a file? Thanks!

HyukjinKwon:
@summerela In Spark 2.0 this was solved and became unlimited. I think anyway this might have to be fixed in this library as well. There wouldn't be a way to set this manually for now. If you are urgent, you might be able to do this like the below one:

```scala
val path = "path-to-file"
val baseRDD = sc.textFile(path)
val csvRDD = baseRDD.filter(! _..startsWith("#"))
val df = new CsvParser().csvRdd(sqlContext, csvRDD)
```

summerela:
Thank you @HyukjinKwon! That'll do the trick for now :)

apurva-sharma:
Does anyone know if this change was ported over in spark 2.2?
I run into the same issue again, spark csv source only seems to take a path as an argument.

PR #132, opened 2015-08-20
Printing content of records when failing or dropping

### LINKED ISSUE #131: FAILFAST mode does not show which line it failed on
It would be good if the FAILFAST exception gave the line number or the contents of the line it failed on. Currently I get something like - 
    
    Malformed line in FAILFAST mode: [Ljava.lang.String;@5f16e0c

Looking at https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/CsvRelation.scala#L109 it seems that this is because the entire $tokens array is being printed instead of using something like Array.toString(). 

Are there any plans of fixing this anytime soon?

falaki:
Constructing printable string when logging content of malformed lines.

closes #131


brkyvz:
LGTM

PR #143, opened 2015-09-09
Build changes / fixes for 1.5.0 final release

JoshRosen:
This pull request contains build changes to reflect the final Spark 1.5.0 release.

I also changed the compilation Spark version to 1.5.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
       scala: 2.11.6
       env: TEST_SPARK_VERSION="1.5.0"
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION -Dspark.testVersion=$TEST_SPARK_VERSION coverage test
+  - sbt -Dspark.testVersion=$TEST_SPARK_VERSION ++$TRAVIS_SCALA_VERSION coverage test
   - sbt ++$TRAVIS_SCALA_VERSION scalastyle
 after_success:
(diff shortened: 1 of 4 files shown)


JoshRosen:
Huh, so it looks like switching the build to test against 1.4.1 fixes the weird compilation issue. I'm going to take a look at the classpaths to see if I can figure out why this would be the case.

JoshRosen:
@falaki, I've cleaned up this PR to include the 1.4.1 test change that we discussed. If this looks okay to you, let me know and we can merge it with the new script.

falaki:
Thanks. Merging it now.

JoshRosen:
@falaki, I've updated this to build against 1.5.0 as the compile-time Spark version; PTAL.

falaki:
Thanks!

barrybecker4:
I see that the 1.5.0 release of spark-csv builds against spark 1.6.0 in the sbt file. Is there a reason why you did not build against 1.6.2? If I use version 1.6.2 version of spark in my application, can I still use the released 1.5.0 version of spark-csv, or should I build my own version where I specify 1.6.2 spark? Will it matter?

falaki:
@barrybecker4 1.6.2 should be API compatible with 1.6.0. That means you should be able to use saprk-csv 1.5.0 with any 1.6.x (or even older versions of spark). Please open a new issue if you encounter problems.

PR #148, opened 2015-09-15
Enforce more Scalastyle rules and add MiMa binary compatibility checks

JoshRosen:
This patch adds MiMa binary compatibility checks and enables a number of Scalastyle rules (after fixing the existing violations).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -86,7 +86,11 @@ mimaDefaultSettings ++ Seq(
   binaryIssueFilters ++= Seq(
     // These classes are not intended to be public interfaces:
     ProblemFilters.excludePackage("com.databricks.spark.csv.CsvRelation"),
-    ProblemFilters.excludePackage("com.databricks.spark.csv.util.InferSchema")
+    ProblemFilters.excludePackage("com.databricks.spark.csv.util.InferSchema"),
+    ProblemFilters.excludePackage("com.databricks.spark.sql.readers"),
+    // We allowed the private `CsvRelation` type to leak into the public method signature:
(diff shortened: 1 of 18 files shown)


build.sbt
```diff
@@ -73,3 +73,25 @@ ScoverageSbtPlugin.ScoverageKeys.coverageHighlighting := {
...
+
+mimaDefaultSettings ++ Seq(
+  previousArtifact := Some("com.databricks" %% "spark-csv" % "1.2.0"),
+  binaryIssueFilters ++= Seq(
+    // These classes are not intended to be public interfaces:
+    ProblemFilters.excludePackage("com.databricks.spark.csv.CsvRelation"),
+    ProblemFilters.excludePackage("com.databricks.spark.csv.util.InferSchema"),
+    ProblemFilters.excludePackage("com.databricks.spark.sql.readers"),
+    // We allowed the private `CsvRelation` type to leak into the public method signature:
+    ProblemFilters.exclude[IncompatibleResultTypeProblem](
```
JoshRosen: This is an example of why it's important to enforce explicit return types in public methods: the inferred type was CsvRelation which, as far as I know, was not intended to be a public class.

falaki:
Why moving ```CsvReader``` to ```readers/readers```?

JoshRosen:
@falaki, its package did not correspond to its filename / location.

falaki:
Code changes LGTM. Please take a look at the test compilation failure in ```ReaderSuite.scala```

JoshRosen:
Updated, so this is now passing tests.

Issue #150, opened 2015-09-17
Assembly build not working


syepes:
Hello,

I am trying to build spark-csv with all its dependencies, but I am getting the following error.
Anyone know what is causing this issue?

```
# sbt/sbt +assembly
...
[info] ScalaTest
[info] Run completed in 37 seconds, 970 milliseconds.
[info] Total number of tests run: 93
[info] Suites: completed 6, aborted 0
[info] Tests: succeeded 93, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 96, Failed 0, Errors 0, Passed 96
java.lang.RuntimeException: Please add any Spark dependencies by supplying the sparkVersion and sparkComponents. Please remove: org.apache.spark:spark-core:1.5.0:test
        at scala.sys.package$.error(package.scala:27)
        at sbtsparkpackage.SparkPackagePlugin$$anonfun$12$$anonfun$apply$1.apply(SparkPackagePlugin.scala:107)
        at sbtsparkpackage.SparkPackagePlugin$$anonfun$12$$anonfun$apply$1.apply(SparkPackagePlugin.scala:95)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.immutable.List.foreach(List.scala:318)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
        at scala.collection.AbstractTraversable.map(Traversable.scala:105)
        at sbtsparkpackage.SparkPackagePlugin$$anonfun$12.apply(SparkPackagePlugin.scala:95)
        at sbtsparkpackage.SparkPackagePlugin$$anonfun$12.apply(SparkPackagePlugin.scala:93)
        at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
        at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
        at sbt.std.Transform$$anon$4.work(System.scala:63)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
        at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
        at sbt.Execute.work(Execute.scala:235)
[...]

PR #151, opened 2015-09-18
Fix SBT assembly command

JoshRosen:
This fixes the SBT assembly command and updates the Travis build to test it. See #150.


brkyvz:
LGTM

PR #153, opened 2015-09-22
the data was not expected if delimiter is '|'

azuryyu:
I fixed two issues here:

1) if CSV data was delimited by '|', then the data was separated by each character.
2) if the column header has non-alphabet and not-digit, such as '(' , space etc,  save as ORC on Spark was failed.

so, I just:
1) if delimiter is '|', then changed to '\u0003' internally.
2) normalize the column header to keep only alphabet and digit, replace other characters with '_'

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -45,7 +45,7 @@ class CsvParser extends Serializable {
   }
 
   def withDelimiter(delimiter: Character): CsvParser = {
-    this.delimiter = delimiter
+    this.delimiter = if (delimiter == '|') '\u0003' else delimiter
     this
   }
 
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -126,6 +126,37 @@ case class CsvRelation protected[spark] (
     }
   }
 
+  private def normalizeHeader(header : String, csvFormat : CSVFormat) = {
+    val newHeader = new StringBuilder
+    val builder   = new StringBuilder
+
+    header.toLowerCase.split(csvFormat.getDelimiter.toString, -1).foreach { head => 
+      var lastChar = ' '      
+      builder.clear
+      
+      val head_ = head.trim
+    
+      for(i <- 0 until head_.length) {
+        if (('a' to 'z' contains head_(i)) ||
+            ('0' to '9' contains head_(i)) ||
+            head_(i) == '_') {
+          builder append head_(i)
+          lastChar = head_(i)
+        }
+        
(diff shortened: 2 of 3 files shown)


HyukjinKwon:
For the first issue you met, I tested this before like this.

```scala
 var df = sqlContext.read
                .option("delimiter","|")
                .load("lineitem.tbl")
```

And this looked fine for me. The data should be well-formed since the file was generated from `dbgen` downloaded from TPC-H official site.

azuryyu:
Can you share your test data? or where to download? I want to test again.

HyukjinKwon:
You can follow up this link (data generation part).

https://husnusensoy.wordpress.com/2010/10/22/create-your-own-oracle-tpc-h-playground-on-linux/

HyukjinKwon:
For the second issue you met, I think it might be better to create an issue to check if this is an issue and if so, to discuss the approach to deal with it.

azuryyu:
OK, I’ll test it recently.

HyukjinKwon:
Does that mean closing this for now?

azuryyu:
No, don’t close.

I mean I’ll test it to confirm this issue. but I mostly sure it’s a really issue.

azuryyu:
for #2:

You guys can reproduce as follow:

0) add  --packages "com.databricks:spark-csv_2.10:1.2.0" in spark-shell

1) prepare test data:[only two rows, you can place '+' or space or '(' etc in the header except '_' ]
  my header|my+header2
  text1|text2

2) put on HDFS: /tmp/csv_test
3) <Hit exception>
val df = sqlContext.read
                              .option("header", "true")
                              .option("delimiter", "|")
                              .load("/tmp/csv_test")
df.write.orc("/tmp/csv_test_output")

pls comment here to show me you can reproduce or not.

azuryyu:
for issue 1:

It mostly because of my incorrect delimiter parameter passed. I cannot reproduce now. big pardon.

HyukjinKwon:
I see. For the second issue, I agree that it is problematic.

However, I just wonder if it is a correct to way convert the original names in the header to other names.

As far as I know, those special characters are not SQL standard ones (and for this reason the error occurs you said). I feel like users should not use such fields names rather than this library supports them by converting them into other names.

Also, if this library must support this, I am not too sure if the `_` is a correct character to represent the special characters. I would rather encode them to represent the original characters.

As far as I know, it is generally true that a library (or plugin) follows the rules of the main program such as MongoDB, Elasticsearch, Solr and etc. In my case, I have seen, in some projects, they encode some special characters when they must use them but have not seen replacing it to another. 

Maybe other guys think differently. 
Would you like to close this PR and open an new issue to discuss for this affair?

azuryyu:
I close it and open a new PR to discuss issue #2.

azuryyu:
I closed and created 
https://github.com/databricks/spark-csv/pull/202 <https://github.com/databricks/spark-csv/pull/202>

PR #154, opened 2015-09-22
TypeCast double and float according to your Locale if default cast fails

gasparms:
In some regions decimal types have comma instead dot.  Currently spark-csv cast to Double or Float using .toDouble or toFloat scala functions. 

If we have double or float types with comma (and the delimiter is another character) we could try to use the java.text.NumberFormat and get the default Locale.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -16,7 +16,8 @@
 package com.databricks.spark.csv.util
 
 import java.math.BigDecimal
-import java.sql.{Timestamp, Date}
+import java.sql.{Date, Timestamp}
+import java.text.NumberFormat
 import java.util.Locale
 
 import org.apache.spark.sql.types._
@@ -52,9 +53,9 @@ object TypeCast {
         case _: IntegerType => datum.toInt
         case _: LongType => datum.toLong
         case _: FloatType => Try(datum.toFloat)
-          .getOrElse(java.text.NumberFormat.getInstance(Locale.getDefault).parse(datum).floatValue())
+          .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).floatValue())
         case _: DoubleType => Try(datum.toDouble)
-          .getOrElse(java.text.NumberFormat.getInstance(Locale.getDefault).parse(datum).doubleValue())
+          .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).doubleValue())
         case _: BooleanType => datum.toBoolean
         case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
         // TODO(hossein): would be good to support other common timestamp formats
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -35,7 +38,8 @@ object TypeCast {
    * @param datum string value
    * @param castType SparkSQL type
    */
-  private[csv] def castTo(datum: String, castType: DataType, nullable: Boolean = true): Any = {
+  private[csv] def castTo(datum: String, castType: DataType, nullable: Boolean = true,
+                          locale: Locale = Locale.getDefault): Any = {
```
falaki: This does not seem to be passed anywhere. That is who is going to pass ```local``` to castTo?

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -44,8 +48,10 @@ object TypeCast {
         case _: ShortType => datum.toShort
         case _: IntegerType => datum.toInt
         case _: LongType => datum.toLong
-        case _: FloatType => datum.toFloat
-        case _: DoubleType => datum.toDouble
+        case _: FloatType => Try(datum.toFloat)
+          .getOrElse(java.text.NumberFormat.getInstance(locale).parse(datum).floatValue())
```
falaki: Can't we just use ```NumberFormat.parse``` by default and avoid catching the exception? Is there a performance penalty?

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -35,7 +38,8 @@ object TypeCast {
    * @param datum string value
    * @param castType SparkSQL type
    */
-  private[csv] def castTo(datum: String, castType: DataType, nullable: Boolean = true): Any = {
+  private[csv] def castTo(datum: String, castType: DataType, nullable: Boolean = true,
```
falaki: Unfortunately, MiMa checker fails on this although it is ```private[csv]```. Please Put a style checker exclusion clause around this line to fix the MiMa failure.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
gasparms: You are right,  I add this parameter just for testing, because if I use default Locale  in mi local environment test assert(TypeCast.castTo("1,00", FloatType) == 1.0) pass but travis fails because its locale its different.  What do you propose to solve this?

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
gasparms: I think we can't, Users must have a choice for parse double and floats with dots or commas.  The most appropriate way is passing a parameter and the code will be clear. Do you like this option?

Other option is use the locale capturing the exception.. obviously there is a small perfomance penalty.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
gasparms: I don't know MiMa, when we decides the best option, I'll tackle this task if necessary.

gasparms:
Now is working using the default locale when the exception is throwed.  If you want to use a parameter, tell me and I'll change the implementation.

falaki:
Looks good to me. Please rebase and I will merge it after another found of tests. Thanks.

gasparms:
Done

Issue #159, opened 2015-10-04
Can we setup SparkSQL to allow for CREATE TABLE using csv vs using com.databricks.spark.csv 


rlgarris:

falaki:
I think this can by done by registering an alias that maps "csv" to "com.databricks.spark.csv"
@yhuai what do you think?

brkyvz:
I remember someone was planning to work on this on the mailing list. There
might be a JIRA for it already.

rxin:
It's already in. https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/sources/interfaces.scala#L63

Also need to register a meta file for the package.

falaki:
Cool. I update spark-csv to use this feature now.

PR #160, opened 2015-10-04
Adding a short name for spark-csv pacakge

falaki:
Users can now use spark-csv with following simpler query:

```
CREATE TEMPORARY TABLE myTable 
USING csv OPTIONS (path "/tmp/data", header "true")
```
No need to specify "com.databricks.spark.csv" as data source name

closes Issue #159

### DIFF
--- a/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
+++ b/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
@@ -1,1 +1,1 @@
-com.databricks.spark.csv.DefaultSource
+com.databricks.spark.csv.DefaultSource15
--- a/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
+++ b/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
@@ -1,0 +1,1 @@
+com.databricks.spark.csv.DefaultSource
--- a/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
@@ -36,11 +36,6 @@ class DefaultSource
   }
 
   /**
-   * Short alias for spark-csv data source.
-   */
-  override def shortName(): String = "csv"
-
-  /**
    * Creates a new relation for data store in CSV given parameters.
    * Parameters have to include 'path' and optionally 'delimiter', 'quote', and 'header'
    */
--- a/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
@@ -28,8 +28,7 @@ import com.databricks.spark.csv.util.{ParserLibs, TextFile, TypeCast}
 class DefaultSource
   extends RelationProvider
   with SchemaRelationProvider
-  with CreatableRelationProvider
-  with DataSourceRegister {
+  with CreatableRelationProvider {
 
   private def checkPath(parameters: Map[String, String]): String = {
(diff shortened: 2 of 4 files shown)


falaki:
ping @rxin

JoshRosen:
Looks like this will break compatibility with pre-1.5.0 versions of Spark:

```
[info] Exception encountered when attempting to run a suite with class name: com.databricks.spark.csv.CsvSuite *** ABORTED ***
[info]   java.lang.NoClassDefFoundError: org/apache/spark/sql/sources/DataSourceRegister
[info]   at java.lang.ClassLoader.defineClass1(Native Method)
[info]   at java.lang.ClassLoader.defineClass(ClassLoader.java:643)
[info]   at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[info]   at java.net.URLClassLoader.defineClass(URLClassLoader.java:277)
[info]   at java.net.URLClassLoader.access$000(URLClassLoader.java:73)
[info]   at java.net.URLClassLoader$1.run(URLClassLoader.java:212)
[info]   at java.security.AccessController.doPrivileged(Native Method)
[info]   at java.net.URLClassLoader.findClass(URLClassLoader.java:205)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:323)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:316)
```

falaki:
Ah, not too unexpected. Suggestions?

rxin:
Does this only break source compatibility, or does it also break binary
compatibility?

JoshRosen:
@rxin, it depends on what you mean exactly, but I think this is a binary compatibility issue: the `DataSourceRegister` trait / interface is not present in pre-1.5.0 versions of Spark, so this package's `DefaultSource` is going to fail during classloading on pre-1.5 versions of Spark. It's also a source-compatibility issue in the sense that this library itself will no longer be compilable from source when using earlier versions of Spark as the compile-time dependencies, but that's only an issue which affects `spark-csv` developers, not users.

I don't think that there are any easy reflection tricks to work around this. If we're open to crazy hacks, we might be able to just copy the `DataSourceRegister` classes into this project and use some classloader tricks to fall back on loading those when we're running on pre-1.5.0 Spark, but this seems brittle and overly-complicated.

The easiest approach is probably to release a new `spark-csv` version which only supports 1.5.0+. If we do this, there is a question of whether we want to backport bugfix patches to pre 1.5-compatible versions, etc. Let's discuss more later.

HyukjinKwon:
Could I leave a comment? I recently worked around elasticsearch-hadoop (https://github.com/elastic/elasticsearch-hadoop) and I saw they had the similar issue and they resolved by decoupling two `DefaultSource` (because the `DefaultSource` for Spark 1.5 is lazily loaded). I think this can be done in a similar way like they did.

They made a `DefaultSource15` like this.
https://github.com/elastic/elasticsearch-hadoop/blob/master/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource15.scala

and support different filters (for upper-versions) like this.
https://github.com/elastic/elasticsearch-hadoop/blob/master/spark/sql-13/src/main/scala/org/elasticsearch/spark/sql/DefaultSource.scala

JoshRosen:
@HyukjinKwon, thanks a bunch for that link; that's a really clever trick. I'll try that approach in `spark-redshift`, since it seem easier than what I had proposed over there: https://github.com/databricks/spark-redshift/pull/105

falaki:
@HyukjinKwon Thanks for the suggestion. Will try it.

falaki:
@JoshRosen please take a look at this. It seems we got it passing thanks to @HyukjinKwon suggestion.

HyukjinKwon:
Cool!

PR #165, opened 2015-10-14
Nullable quote character

### LINKED ISSUE #89: Support for disabling CSV quote
I have some tab separated files that may contain all sorts of other characters. I do not want to interpret any character as the quote character, however currently I can't find a way to set it to null, which by the apache commons-csv documentation would disable quoting:

https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#withQuote(char)

"quoteChar - the quoteChar character, use null to disable"

I'd submit a pull request, but I can't find how the options map gets propagated to the data format package. I've tried:

```scala
sqlContext.createDataFrame(rowRDD, outputSchema)
      .write.format(Constants.CSVDataSource)
      .option("quote", null)
```
But that causes a `NullPointerException`. Can we expose this somehow through `.option("noQuote", "true")` or similar?

jamesblau:
This pull request adds to spark-csv the functionality to provide a null for the quote character, in the event that no quoting is desired. This is necessary to interoperate with libraries such as Scalding which use unquoted formats by default.

Since the uniVocity parser uses a non-nullable char for the quoting character, a null byte is substituted in CsvRelation.scala if null was provided and uniVocity used, just as is already done for commentChar.

A test is also added using a comma-separated file with unbalanced double-quote characters.

Author: James Blau <james@tresata.com>

Closes #89 from tresata-opensource/feat-nullable-quotation-character.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -36,7 +36,7 @@ case class CsvRelation protected[spark] (
     location: Option[String],
     useHeader: Boolean,
     delimiter: Char,
-    quote: Char,
+    quote: Character,
     escape: Character,
     comment: Character,
     parseMode: String,
@@ -143,7 +143,8 @@ case class CsvRelation protected[spark] (
       val firstRow = if (ParserLibs.isUnivocityLib(parserLib)) {
         val escapeVal = if (escape == null) '\\' else escape.charValue()
         val commentChar: Char = if (comment == null) '\0' else comment
-        new LineCsvReader(fieldSep = delimiter, quote = quote, escape = escapeVal,
+        val quoteChar: Char = if (quote == null) '\0' else quote
+        new LineCsvReader(fieldSep = delimiter, quote = quoteChar, escape = escapeVal,
           commentMarker = commentChar).parseLine(firstLine)
       } else {
         val csvFormat = CSVFormat.DEFAULT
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -31,7 +31,7 @@ package object csv {
         filePath: String,
         useHeader: Boolean = true,
         delimiter: Char = ',',
-        quote: Char = '"',
```
falaki: Since this is an exposed API, can you keep it a ```Char``` cast it before passing to ```CsvRelation```

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -143,7 +143,8 @@ case class CsvRelation protected[spark] (
       val firstRow = if (ParserLibs.isUnivocityLib(parserLib)) {
         val escapeVal = if (escape == null) '\\' else escape.charValue()
         val commentChar: Char = if (comment == null) '\0' else comment
-        new LineCsvReader(fieldSep = delimiter, quote = quote, escape = escapeVal,
+        val quoteChar: Char = if (quote == null) '\0' else quote
+        new LineCsvReader(fieldSep = delimiter, quote = quoteChar, escape = escapeVal,
```
falaki: Style nit:
```
new LineCsvReader(
  fieldSep = delimiter, 
  quote = quoteChar, 
  escape = escapeVal,
  commentMarker = commentChar).parseLine(firstLine)
```

falaki:
Thanks for submitting this. Please address the binary compatibility error.

PR #167, opened 2015-10-16
Added defaultCsvFormat member to csv/package to use CSVFormat.DEFAULT…

murrayg:
Added defaultCsvFormat member to csv/package to use CSVFormat.DEFAULT with system line separator.

Spark-csv uses `org.apache.commons.csv.CSVFormat.DEFAULT` which uses line separator `\r\n`.  On UNIX systems, this results in the header line in output files to show up with a `^M` line ending.  I.e., the output looks like
```
Id      Name    City    Name_1  CustId  State^M
38228   Kathy   Callaway        Kathryn 38228   CA
45639   Ethel   Brownsville     Ethel   45639   AK
```
with the spurious `^M` on the header line. 

This fork addresses this issue by substituting a `defaultCsvFormat` member that sets the line separator to the system value.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/package.scala
+++ b/src/main/scala/com/databricks/spark/csv/package.scala
@@ -23,6 +23,8 @@ import com.databricks.spark.csv.util.TextFile
 
 package object csv {
 
+  val defaultCsvFormat = CSVFormat.DEFAULT.withRecordSeparator(System.lineSeparator())
+
   /**
    * Adds a method, `csvFile`, to SQLContext that allows reading CSV data.
    */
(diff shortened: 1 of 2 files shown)


falaki:
@murrayg This does not compile on Java 6.

murrayg:
I've pushed a fix to my fork:

-  val defaultCsvFormat = CSVFormat.DEFAULT.withRecordSeparator(System.lineSeparator())
+  val defaultCsvFormat = CSVFormat.DEFAULT.withRecordSeparator(System.getProperty("line.separator"))

Is there some way to try this again or do I need to do another pull request?

Thanks,
Glenn

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -23,6 +23,8 @@ import com.databricks.spark.csv.util.TextFile
 
 package object csv {
 
+  val defaultCsvFormat = CSVFormat.DEFAULT.withRecordSeparator(System.getProperty("line.separator"))
```
falaki: ```getProperty``` may return null if the property is not set. Please pass it a second argument as default value.

falaki:
@murrayg Thanks for the update. I left one minor comment.

Issue #166, opened 2015-10-15
Support for pruned scan for required fields.


HyukjinKwon:
As I look through the codes, it looks it reads all the data without pruning by required fields, which i think might improve its performance in particular when the number of requested columns is few but the schema is huge (by io cost and etc.).

I understand that the parser itself parses line by line though.
Is this intentionally unimplemented for some reasons?

falaki:
We always appreciate community contribution.

JoshRosen:
Would this make a performance impact for CSV? Given a row-oriented file format like CSV, how do you save IO / network traffic by skipping reads of certain columns? I can see how column-pruning makes sense for column-oriented file formats, but I don't see an obvious way to do this for CSV.

HyukjinKwon:
There wouldn't be tremendous improvement though. I can see JSONRelatiom in Spark supports this by JacksonParser which supports skip by token.

If we know which position (rather index of fields) to read and parse, I think it can skip token by token (just like JacsonParser) without type checking and read but just move offset and position of the current stream.

HyukjinKwon:
I will close this isuue since this issue is resolved.

PR #169, opened 2015-10-20
Support for pruned scan for required fields.

HyukjinKwon:
https://github.com/databricks/spark-csv/issues/166

Firstly,
I tested the pure iteration time of produced RDD from `buildScan()` with `lineitem` table in TPC-H.
On local machine, I set the number of 1,000,000 rows and measured the time.

The result was below. (The total number of columns was 17)

|Scan Type|Required Fields|Elapsed Time|
|--------|-------|-------|
|`PrunedScan`|`C0`|0.106|
|`PrunedScan`|`C0, C1, C2, C3, C4, C5, C6`|0.287|
|`PrunedScan`|`* (C0 - C16)`|0.505|
|`TableScan`|`C0`|0.536|
|`TableScan`|`C0, C1, C2, C3, C4, C5, C6`|0.502|
|`TableScan`|`* (C0 - C16)`|0.489|

Secondly, I didn't add test codes separately since the original test codes cover this.

Could you review this please?


JoshRosen:
Cool! I'll take a closer look at this tomorrow.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,58 @@ case class CsvRelation protected[spark] (
...
+          None
+        } else if (failFast && schemaFields.length != tokens.size) {
+          throw new RuntimeException(s"Malformed line in FAILFAST mode: ${tokens.mkString(",")}")
+        } else {
+          val indexSafeTokens = if (permissive && schemaFields.length != tokens.size) {
+            tokens ++ new Array[String](schemaFields.length - tokens.length)
+          } else {
+            tokens
+          }
+          var index: Int = 0
```
JoshRosen: Minor code simplification: I think that `index` and `subIndex` can be declared inside of the `try` block in order to use two fewer lines of code.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,58 @@ case class CsvRelation protected[spark] (
...
+          var subIndex: Int = 0
+          val rowArray = new Array[Any](requiredIndices.length)
```
JoshRosen: If you're interested in leveraging mutability as a performance optimization, then I think you might be able to re-use a single mutable array rather than creating a new one for each row.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,58 @@ case class CsvRelation protected[spark] (
...
+        .foreach(pair => requiredIndices.update(requiredFields.indexOf(pair._1), pair._2))
+
+      tokenRdd(schemaFields.map(_.name)).flatMap { tokens =>
+        if (dropMalformed && schemaFields.length != tokens.size) {
+          logger.warn(s"Dropping malformed line: ${tokens.mkString(",")}")
+          None
+        } else if (failFast && schemaFields.length != tokens.size) {
+          throw new RuntimeException(s"Malformed line in FAILFAST mode: ${tokens.mkString(",")}")
+        } else {
+          val indexSafeTokens = if (permissive && schemaFields.length != tokens.size) {
```
JoshRosen: If `tokens` contains fewer columns than expected and we're _not_ in permissive mode, then should we throw a more explicit error rather than allowing an ArrayIndexOutOfBounds to occur?

JoshRosen: Oh, nevermind, I see that this is covered by the `if (failFast && schemaFields.length != tokens.size)` case above.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
JoshRosen: On the other hand, the non-pruned-scan codepath didn't do this, so we should leave this optimization to a followup.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
HyukjinKwon: oh right thanks!

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
HyukjinKwon: I though the use of a mutable array decreases performance.
Thanks, I will update them soon.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
 
+  override def buildScan(requiredColumns: Array[String]): RDD[Row] = {
+    val schemaFields = schema.fields
+    val requiredFields = StructType(requiredColumns.map(schema(_))).fields
+    val isTableScan = requiredColumns.isEmpty || schemaFields.deep == requiredFields.deep
+    if (isTableScan) {
+      buildScan
+    } else {
+      val requiredIndices = new Array[Int](requiredFields.length)
+      schemaFields.zipWithIndex.filter(pair => requiredFields.contains(pair._1))
```
falaki: Nit: would be more readable if you match the tuple:
```scala
schemaFields.zipWithIndex.filter { 
  case (field, _) requiredFields.contains(field)
}
```

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
+        .foreach(pair => requiredIndices.update(requiredFields.indexOf(pair._1), pair._2))
```
falaki: Same here. For better readability please match the tuple inside foreach: ```foreach { case(field, index) => ... }```

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
+
+      val rowArray = new Array[Any](requiredIndices.length)
+      tokenRdd(schemaFields.map(_.name)).flatMap { tokens =>
+        if (dropMalformed && schemaFields.length != tokens.size) {
+          logger.warn(s"Dropping malformed line: ${tokens.mkString(",")}")
```
falaki: Would be better to log the actual delimiter:
```scala
logger.warn(s"Dropping malformed line: ${tokens.mkString(delimiter)}")
```

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
+          None
+        } else if (failFast && schemaFields.length != tokens.size) {
+          throw new RuntimeException(s"Malformed line in FAILFAST mode: ${tokens.mkString(",")}")
```
falaki: Same here.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
+          } else {
+            tokens
+          }
+          try {
+            var index: Int = 0
+            var subIndex: Int = 0
+            while (subIndex < requiredIndices.length) {
+              index = requiredIndices(subIndex)
+              val field = schemaFields(index)
+              rowArray(subIndex) = TypeCast.castTo(indexSafeTokens(index), field.dataType,
```
falaki: style nit:
```scala
rowArray(subIndex) = TypeCast.castTo(
    indexSafeTokens(index), 
    field.dataType,
    field.nullable,
    treatEmptyValuesAsNulls)
```

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
+                field.nullable, treatEmptyValuesAsNulls)
+              subIndex = subIndex + 1
+            }
+            Some(Row.fromSeq(rowArray))
+          } catch {
+            case nfe: java.lang.NumberFormatException if dropMalformed =>
+              logger.warn("Number format exception. " +
+                s"Dropping malformed line: ${tokens.mkString(",")}")
```
falaki: Same here with original delimiter.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
...
+              None
+            case pe: java.text.ParseException =>
+              logger.warn("Parse Exception. " +
+                s"Dropping malformed line: ${tokens.mkString(",")}")
```
falaki: and here.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -136,6 +136,56 @@ case class CsvRelation protected[spark] (
     }
   }
 
+  override def buildScan(requiredColumns: Array[String]): RDD[Row] = {
```
falaki: Would you please add some comments to this method and explain the general logic? This will help other developers quickly understand what is going on when they want to contribute to this code.

falaki:
@HyukjinKwon Thanks for adding this feature. I left some comments, mostly regarding style and readability.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
HyukjinKwon: Right! I will add comments like the ones in `org.apache.spark.sql.sources.PrunedScan` at Spark
Also, let me update them (below) soon.

HyukjinKwon:
@falaki Thanks for your advices. I updated them.
Also, I changed `requiredIndices.update(requiredFields.indexOf(field), index)` to `requiredIndices(requiredFields.indexOf(field)) = index`,
and `java.text.ParseException` to `java.text.ParseException if dropMalformed =>` which I think I removed  by mistake while testing.

PR #180, opened 2015-11-02
Fixing incorrect options value in README.md

xguo27:
Hi:

I noticed a small typo in the readme.md which makes the example throws syntax error.

Here is the change.

Thanks!


falaki:
Thanks for the correction.

PR #181, opened 2015-11-03
Fix syntax errors in Python examples

sryza:


falaki:
Thanks for the correction.

PR #185, opened 2015-11-10
Add escape feature to documentation

thoralf-gutierrez:
The documentation was missing the escape character that's available through `CsvParser.withEscape`


falaki:
@thoralf-gutierrez thanks for sending this. Would you please update it. It is getting merge conflicts right now.

thoralf-gutierrez:
I just rebased, it should be good now :+1:

PR #186, opened 2015-11-10
Add Min partitions to CSV Reading

hntd187:
I added the ability to specify a minimum amount of partitions, but my style obviously is conflicting with your style, mine passes your scala style, but could you give me some more information about your style so I can reformat this correctly? I'm using eclipse.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -127,8 +127,7 @@ class CsvParser extends Serializable {
       ignoreTrailingWhiteSpace,
       treatEmptyValuesAsNulls,
       schema,
-      inferSchema,
-      minPartitions)(sqlContext)
+      inferSchema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
   }
 
@@ -147,8 +146,7 @@ class CsvParser extends Serializable {
       ignoreTrailingWhiteSpace,
       treatEmptyValuesAsNulls,
       schema,
-      inferSchema,
-      minPartitions)(sqlContext)
+      inferSchema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
   }
 }
--- a/src/main/scala/com/databricks/spark/csv/CsvParser.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvParser.scala
@@ -15,14 +15,12 @@
  */
 package com.databricks.spark.csv
 
-
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.{DataFrame, SQLContext}
+import org.apache.spark.sql.{ DataFrame, SQLContext }
 import org.apache.spark.sql.types.StructType
-import com.databricks.spark.csv.util.{ParserLibs, ParseModes, TextFile}
+import com.databricks.spark.csv.util.{ ParserLibs, ParseModes, TextFile }
 
-/**
- * A collection of static functions for working with CSV files in Spark SQL
+/** A collection of static functions for working with CSV files in Spark SQL
  */
 class CsvParser extends Serializable {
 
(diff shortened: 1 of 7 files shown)


hntd187:
I am unsure why this is failing, could someone provide some insights?

falaki:
@hntd187 why do you think this should be a parameter of the data source and not SparkSQL?

hntd187:
@falaki It seemed obvious to mimic the way textFile has its input, but if you feel it shouldn't be here or should be somewhere else I trust your judgement more than mine on this matter.

falaki:
Yes. Would you suggest this to SparkSQL (on the mailing list)?

hntd187:
@falaki could you clarify what you are suggesting I pose to the mailing group?

falaki:
I meant propose this change to the Spark community. For example, you can file a JIRA or post your suggestion to spark-dev mailing list.

koertkuipers:
Why does this require a change for spark-sql? It doesnt conflict with any existing api?

falaki:
This is basically out of scope for a data source.

koertkuipers:
You mean you would like it to be some interface so spark-sql can control it in a more generic way? So it can become part of what the planner does perhaps?

koertkuipers:
I am doing:
```
val rdd = sqlContext.read
      .option(..., ...)
      .option("minPartitions", 1000)
      .load("/some/path")
      .rdd
```
Basically i used this for RDDs. If it does not go into the this data source then i am unsure where it would go in my example?

falaki:
You can simply do this:
```scala
val rdd = sqlContext.read
      .option(..., ...)
      .load("/some/path")
      .repartition(1000)
      .rdd
```

koertkuipers:
its very different currently.

minPartitions pushes the logic of reducing partition size to increase
number of partitions into hadoop's FileFormat. it is very efficient.
DataFrame.repartition to a larger number of partitions causes an expensive
shuffle. this shuffle is exactly what i want to avoid.

also if i read from a data source that already splits to 1000+ partitions
using hadoop's blocksize, minPartitions does nothing (it only increases
partitions), while DataFrame.repartition will reduce number of partitions
(efficiently, no shuffle needed).

HyukjinKwon:
@koertkuipers I think it is to deal with some narrow use-cases. We might be able to do this by `colerase()` without shuffle. 

Although it would not work identically with `minPartitions`, do you think a lot of guys would use this `minPartitions` option, while this can be done with workarounds such as `colerase()` and `repartition()` which are already pretty much being used?

For me, now there are pretty a lot of options and I feel like we should add some options only if a lot of guys need this..

koertkuipers:
our main use case is for disk-to-disk algorithms where the initial number
of partitions is too small. by trial and error we have found this to have a
big impact on performance (i am still somewhat confused as to why this is
the case). repartition() or coalesce() in this case would lead to an
expensive shuffle.

it is a narrow use case in that not many people use spark for disk-to-disk
(data that does not fit in memory) i think, yes. but if that is your
yardstick for inclusion then nothing that makes spark a serious big data
platform will make it in anymore...

we are adding minPartitions as a generic option to all hdfs based dataframe
datasources in our in-house versions. its not hard to maintain in a fork.
in fact it would be great if json and parquet datasources were also
properly separated from spark to make maintaining forks of them easier.

thanks! best, koert

hntd187:
Keep me up to date here with what you guys want to actually do with this. I assumed we were going to reject this pull request, but if people actually want to merge it I'll have to resolve the conflicts.

HyukjinKwon:
For `coalesce()`, shuffle is disabled by default [apache/spark/basicOperators.scala#L406-L408](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/basicOperators.scala#L406-L408), although it does not work idencial with `minPartitions`.

That was just my thought. I think we should wait and see for now.

koertkuipers:
coalesce only has shuffle disabled if you reduce number of partitions. if
you want to increase partitions you need shuffle.
minPartitions is to increase number of partitions

HyukjinKwon:
@hntd187 I opened a JIRA for this. Can we maybe close this until we can have a proper conclusion for this?

hntd187:
@HyukjinKwon Sure, feel free to ping me on it. Can you link the JIRA?

HyukjinKwon:
@hntd187 I opened a JIRA here, https://issues.apache.org/jira/browse/SPARK-13184

PR #198, opened 2015-12-02
Exposed parameter for univosity maximum number of characters per column.

adamosloizou:
#### Description
While using the `spark-csv` library with the Univosity parser, we ran into problems when parsing lines with really long columns.
This was due to a predefined limit on the column length that is required by Univosity.

This PR extracts that configuration at the `spark-csv` top level so that it can be changed.

* Added configuration parameter `maxCharsPerColumn` with the same default value of `100000` as it was before.
* Added unit test to cover for the case
* The parameter only applies for the univosity parser and is ignored for the commons parser. This was explained in the README file as well.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/package.scala
+++ b/src/main/scala/com/databricks/spark/csv/package.scala
@@ -52,7 +52,8 @@ package object csv {
         ignoreLeadingWhiteSpace: Boolean = false,
         ignoreTrailingWhiteSpace: Boolean = false,
         charset: String = TextFile.DEFAULT_CHARSET.name(),
-        inferSchema: Boolean = false): DataFrame = {
+        inferSchema: Boolean = false,
+        maxCharsPerColumn: Int = 100000): DataFrame = {
       val csvRelation = CsvRelation(
         () => TextFile.withCharset(sqlContext.sparkContext, filePath, charset),
         location = Some(filePath),
@@ -66,7 +67,8 @@ package object csv {
         ignoreLeadingWhiteSpace = ignoreLeadingWhiteSpace,
         ignoreTrailingWhiteSpace = ignoreTrailingWhiteSpace,
         treatEmptyValuesAsNulls = false,
-        inferCsvSchema = inferSchema)(sqlContext)
+        inferCsvSchema = inferSchema,
+        maxCharsPerColumn = maxCharsPerColumn)(sqlContext)
       sqlContext.baseRelationToDataFrame(csvRelation)
     }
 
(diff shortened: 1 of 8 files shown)


falaki:
@adamosloizou thanks for submitting this. Since you already looked into it, do we have similar limits for the apache commons csv parser?

adamosloizou:
@falaki My pleasure :smile:. I've had a look around the `commons-csv` code a couple of times and couldn't find anything for limiting input. That's why I added the test and comments to state that it's a Univosity only configuration.

falaki:
Would you please look into the test failures? I will do a review.

adamosloizou:
@falaki the failures have to do with binary incompatibilities that are caused by extending the API with the new parameters. The parameters however do come with default values so in Scala the API should be backwards compatible. I don't think we can preserve binary compatibility for this one. What are your thoughts?

falaki:
Yes, please do not modify the signature of `tsvFile` and `csvFile` function.

adamosloizou:
Unfortunately, this is a snowball effect. I can't add the new parameter without breaking the compatibility of some part of the API :worried:

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -77,7 +79,8 @@ package object csv {
         ignoreLeadingWhiteSpace: Boolean = false,
         ignoreTrailingWhiteSpace: Boolean = false,
         charset: String = TextFile.DEFAULT_CHARSET.name(),
-        inferSchema: Boolean = false): DataFrame = {
+        inferSchema: Boolean = false,
+        maxCharsPerColumn: Int = 100000): DataFrame = {
```
falaki: Remove this additional parameter.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -91,7 +94,8 @@ package object csv {
         ignoreLeadingWhiteSpace = ignoreLeadingWhiteSpace,
         ignoreTrailingWhiteSpace = ignoreTrailingWhiteSpace,
         treatEmptyValuesAsNulls = false,
-        inferCsvSchema = inferSchema)(sqlContext)
+        inferCsvSchema = inferSchema,
+        maxCharsPerColumn = maxCharsPerColumn)(sqlContext)
```
falaki: remove this as well.

falaki:
Left two comments on where to remove the change. Instead of specifying default in tsvFile and csvFile, have default value in CsvRelation constructor.

adamosloizou:
@falaki I have incorporated the changes you suggested but that still leaves the `CsvReader`, `LineCsvReader`and the `BulkCsvReader` still with incompatible APIs and those changes are needed.

HyukjinKwon:
Should this maybe re-opened for better testing? The build looks failed as AKAIK Travis CI does not compile the codes all for every commit but tries to compiles only changed ones. During this procedure, it usually throw some compilation errors.
So, if we re-open a new PR, that will trigger the build from scratch which (I think) will resolve the current failure.

falaki:
@adamosloizou this is good to have. In fact we added it to csv data source in spark 2.0. Would you mind rebasing this?

HyukjinKwon:
@adamosloizou Do you mind if I submit a PR based on this if you are currently not working on this?

adamosloizou:
Hi guys, sorry for the very late reply. I understand that due to Spark v2.0 this is no longer needed. Can we safely close this PR? Thanks!

PR #202, opened 2015-12-10
Exception when there are special characters in CSV header

azuryyu:
We definitely hit the exception if there are special characters in CSV header. such as:

header1, header2+test

It's always failed if transform it to ORC, Parquet etc after load CSV.

so two options here:
1) encode the special characters in the header
2) just failed to the user to fix the data


HyukjinKwon:
Hm.. I mean an issue here in https://github.com/databricks/spark-csv/issues :)

azuryyu:
close it and move to issue.

PR #213, opened 2015-12-21
Fix the Java example of custom schemas.

dennishuo:
Needed to use DataTypes.FooType instead of just the class names as the second arg, need to pass the list of StructField as an array, and need to add a Metadata.empty() argument to each StructField constructor.


falaki:
Thanks.

PR #215, opened 2015-12-23
README: Simplify Linking Section

Tarrasch:
In particular, now there is a line one can copy and paste and then edit as needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,19 +10,18 @@ A library for parsing and querying CSV data with Apache Spark, for Spark SQL and
 This library requires Spark 1.3+
 
 ## Linking
-You can link against this library in your program at the following coordiates:
+You can link against this library in your program at the following coordinates:
 
-### Scala 2.10
 ```
 groupId: com.databricks
-artifactId: spark-csv_2.10
+artifactId: spark-csv_2.11 or spark-csv_2.10
 version: 1.3.0
 ```
-### Scala 2.11
+
+For example, if you use `sbt` and scala 2.11. You would add a line like
+
 ```
-groupId: com.databricks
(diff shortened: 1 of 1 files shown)


Tarrasch:
The actual reason I send this PR is that I spent a good frustrating 10 minutes at the least trying to debug why sbt couldn't find the artifact when I typed:

```
libraryDependencies += "com.databricks" %% "spark-csv_2.10" % "1.3.0"
```

Not only didn't I notice the double `%%`, but I still don't understand the difference. =)

README.md
```diff
@@ -10,19 +10,18 @@ A library for parsing and querying CSV data with Apache Spark, for Spark SQL and
...
-artifactId: spark-csv_2.10
+artifactId: spark-csv_2.11 or spark-csv_2.10
 version: 1.3.0
 ```
-### Scala 2.11
+
+For example, if you use `sbt` and scala 2.11. You would add a line like
+
 ```
-groupId: com.databricks
```
falaki: Please do not remove this part. You can add the SBT version (and maybe Maven as well) next to it.

Tarrasch: Sure I'll change it to as you like. But just to avoid confusion: You want there to an explicit mention of the groupId/artifactId/version triple for *both* 2.10 and 2.11. I removed one of them cause I found it very redundant and hard to read.

Is there perhaps a reason I'm not aware of to have them mentioned twice like it is now?

falaki: I prefer more verbose and clear documentation.

HyukjinKwon:
Can we close this for now if you are not going to update this more?

Tarrasch:
Sure.

Issue #219, opened 2015-12-28
Non-consistent result when DROPMALFORMED is enabled and pruned scan is executed.


HyukjinKwon:
DROPMALFORMED parsing mode checks all the data and it only drops only when it fails.

So, in case of `PrunedScan`, this can end up with different results. 

For example, the test below produces the correct results.
```scala
val ageFile = "src/test/resources/ages.csv"
```

```scala
val strictSchema = new StructType(
  Array(
    StructField("Name", StringType, true),
    StructField("Age", IntegerType, true),
    StructField("Height", DoubleType, true)
  )
)

val results = new CsvParser()
  .withSchema(strictSchema)
  .withUseHeader(true)
  .withParserLib(parserLib)
  .withParseMode(ParseModes.DROP_MALFORMED_MODE)
  .csvFile(sqlContext, ageFile)
  .show()
```
Output :
```
+----+---+------+
|Name|Age|Height|
+----+---+------+
| Bob| 10| 180.0|
+----+---+------+
```

This is because it drops two rows here
```
10:17:53.205 WARN com.databricks.spark.csv.CsvRelation$: Number format exception. Dropping malformed line: Joe,20.5,200
10:17:53.206 WARN com.databricks.spark.csv.CsvRelation$: Parse Exception. Dropping malformed line: Tom,30,Dan
```

However, when it turns to `PrunedScan`, this does produces inconsistent results as it does not try to parse `"Age"` and `"Height"`. So, it produces three rows.

```scala
val strictSchema = new StructType(
  Array(
    StructField("Name", StringType, true),
    StructField("Age", IntegerType, true),
    StructField("Height", DoubleType, true)
  )
)

val results = new CsvParser()
  .withSchema(strictSchema)
  .withUseHeader(true)
  .withParserLib(parserLib)
  .withParseMode(ParseModes.DROP_MALFORMED_MODE)
  .csvFile(sqlContext, ageFile)
  .selectExpr("Name")
  .show()
```

Output :
```
+----+
|Name|
+----+
| Bob|
| Joe|
| Tom|
+----+
```

HyukjinKwon:
Fixed in https://github.com/databricks/spark-csv/pull/220

Issue #218, opened 2015-12-24
Unnecessarily castng all values in columns when count is performed.


HyukjinKwon:
When we perform count on `DataFrame`, this produces the plan below.

```
== Physical Plan ==
TungstenAggregate(key=[], functions=[(count(1),mode=Final,isDistinct=false)], output=[count#10L])
+- TungstenExchange SinglePartition, None
   +- TungstenAggregate(key=[], functions=[(count(1),mode=Partial,isDistinct=false)], output=[count#14L])
      +- Scan CsvRelation[] ...
```

But thing is, it performs `TableScan` when `requiredColumns` is empty [CsvRelation.scala#L150](https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/CsvRelation.scala#L150), resulting reading and casting all values in the columns. (The performance test proceeded before in the PR https://github.com/databricks/spark-csv/pull/169)

I was thinking of resolving this by letting this return empty rows or explicitly setting any column of schema since the data in CSV is consistent with schema but I am not too sure if this approach is appropriate and even if this should be fixed.

yhuai:
@HyukjinKwon CSV will read all data anyway. So, you meant creating rows containing all column values instead of creating empty rows, right?

HyukjinKwon:
@yhuai It will read all data but does not have to check all the values in rows and convert (cast) them by pruning scan #169 (merged in CSV). 

I think I should change "read" to one such as convert or cast.

And actually I meant empty rows... although I haven't tested yet, doesn't JSON datasource also produce rows in this way?

yhuai:
@HyukjinKwon Yeah, JSON will produce empty rows if no column is needed.

HyukjinKwon:
Fixed in https://github.com/databricks/spark-csv/pull/220

PR #220, opened 2015-12-28
Inconsistent result for DROPMALFORMED with pruned scan and unnecessarily casting all values when no fields are required

HyukjinKwon:
https://github.com/databricks/spark-csv/issues/218 https://github.com/databricks/spark-csv/issues/219

In this PR, I made the pruned scan try to parse all the values in columns when DROPMALFORMED is enabled and return only required fields.

In addition, I changed the condition for table scan. If required columns are empty, then it just produces empty rows.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -147,18 +147,22 @@ case class CsvRelation protected[spark] (
   override def buildScan(requiredColumns: Array[String]): RDD[Row] = {
     val schemaFields = schema.fields
     val requiredFields = StructType(requiredColumns.map(schema(_))).fields
-    val isTableScan = requiredColumns.isEmpty || schemaFields.deep == requiredFields.deep
-    if (isTableScan) {
+    val shouldTableScan = {
+      val shouldCheckAllTokens = dropMalformed && requiredColumns.isEmpty
+      val isSameSchemaRequired = schemaFields.deep == requiredFields.deep
+      shouldCheckAllTokens || isSameSchemaRequired
+    }
+    if (shouldTableScan) {
       buildScan
     } else {
       val requiredIndices = new Array[Int](requiredFields.length)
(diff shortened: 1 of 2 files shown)


falaki:
Thanks @HyukjinKwon for investigating this. Do you think we should also include FAILFAST mode?

HyukjinKwon:
Ah.. actually it looks fine with FAILFAST mode as it just compares with schema and tokens (which are always read all regardless of pruned scan or table scan).

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -147,18 +147,25 @@ case class CsvRelation protected[spark] (
   override def buildScan(requiredColumns: Array[String]): RDD[Row] = {
     val schemaFields = schema.fields
     val requiredFields = StructType(requiredColumns.map(schema(_))).fields
-    val isTableScan = requiredColumns.isEmpty || schemaFields.deep == requiredFields.deep
-    if (isTableScan) {
+    val shouldTableScan = schemaFields.deep == requiredFields.deep
+    val safeRequiredFields = if (dropMalformed) {
+      // If `dropMalformed` is enabled, then it needs to parse all the values
+      // so that we can decide which row is malformed.
+      requiredFields ++ schemaFields.filterNot(requiredFields.contains(_))
```
falaki: Why not simply using `schemaFields`?

HyukjinKwon:
And.. would you mind if you look through the codes, in particular, the variable names (although it is trivial)?
I am not too sure if the naming looks appropriate.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
HyukjinKwon: Because of the order of columns. The required columns can be in a different order with the original schema.
If you look through the code below, you will notice I slice the row by `take()`.

falaki:
Thanks this looks good to me.

PR #221, opened 2015-12-28
Support for dropping malformed rows for TimestampType and DateType types

HyukjinKwon:
In this PR, I added some logics to catch `IllegalArgumentException`. When trying to parse malformed data to `TimestampType` or `DateType`, it emits the exception with the message below. 

```
java.lang.IllegalArgumentException
	at java.sql.Date.valueOf(Date.java:143)
	at com.databricks.spark.csv.util.TypeCast$.castTo(TypeCast.scala:64)
	at com.databricks.spark.csv.CsvRelation$$anonfun$buildScan$6.apply(CsvRelation.scala:188)
	at com.databricks.spark.csv.CsvRelation$$anonfun$buildScan$6.apply(CsvRelation.scala:169)
	at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:396)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:369)
	...
```

So, because of this, dropping malformed rows does not work properly.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -124,18 +124,13 @@ case class CsvRelation protected[spark] (
           case aiob: ArrayIndexOutOfBoundsException if permissive =>
             (index until schemaFields.length).foreach(ind => rowArray(ind) = null)
             Some(Row.fromSeq(rowArray))
-          case nfe: java.lang.NumberFormatException if dropMalformed =>
+          case _: java.lang.NumberFormatException |
+               _: IllegalArgumentException if dropMalformed =>
             logger.warn("Number format exception. " +
               s"Dropping malformed line: ${tokens.mkString(",")}")
             None
           case pe: java.text.ParseException if dropMalformed =>
             logger.warn("Parse exception. " +
-              s"Dropping malformed line: ${tokens.mkString(",")}")
-            None
-          // When the function `valueOf` in Timestamp or Date fails,
-          // this emits IllegalArgumentException.
-          case iae: IllegalArgumentException if dropMalformed =>
-            logger.warn("Illegal argument exception. " +
               s"Dropping malformed line: ${tokens.mkString(",")}")
             None
         }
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
I am worried of few things.

Firstly, I did not want to add another CSV file and test separately to check this so I just correct original tests and the data. If this is inappropriate, I will get that back and add another test.

Secondly, for me `IllegalArgumentException` and the message `Illegal argument exception` does not sound like it failed during parsing. Would you tell me anything you would prefer if you feel in the same way?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -129,7 +129,13 @@ case class CsvRelation protected[spark] (
               s"Dropping malformed line: ${tokens.mkString(",")}")
             None
           case pe: java.text.ParseException if dropMalformed =>
```
falaki: You can catch both exceptions together:
```Scala
case _ : java.lang.NumberFormatException | _ : IllegalArgumentException =>
```

falaki: Also please add this new exception handling case to `buildScan(requiredColumns: Array[String])` as well.

HyukjinKwon: I definitely will.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -124,13 +124,14 @@ case class CsvRelation protected[spark] (
           case aiob: ArrayIndexOutOfBoundsException if permissive =>
             (index until schemaFields.length).foreach(ind => rowArray(ind) = null)
             Some(Row.fromSeq(rowArray))
-          case nfe: java.lang.NumberFormatException if dropMalformed =>
+          case _: java.lang.NumberFormatException |
+               _: IllegalArgumentException if dropMalformed =>
```
HyukjinKwon: Is this newline appropriate? I cannot find a proper document in [databricks/scala-style-guide](https://github.com/databricks/scala-style-guide).

falaki:
Thanks @HyukjinKwon this looks good.

PR #224, opened 2015-12-29
Add nullValue being respected when parsing CSVs

addisonj:
This change makes it so that we look for a user specified nullValue
through the CSV parsing. This allows for handling CSVs that might use
something else other than an empty string to represent nulls.

It reuses the same flag as CSV saving, `nullValue`. This change should
be non-breaking.

This also pushes this behavior into inferSchema so that inferred schemas
will properly reflect the user given null value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ When reading files the API accepts several options:
 * `inferSchema`: automatically infers column types. It requires one extra pass over the data and is false by default
 * `comment`: skip lines beginning with this character. Default is `"#"`. Disable comments by setting this to `null`.
 * `codec`: compression codec to use when saving to file. Should be the fully qualified name of a class implementing `org.apache.hadoop.io.compress.CompressionCodec`. Defaults to no compression when a codec is not specified.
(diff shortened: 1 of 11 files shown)


addisonj:
Just now noticed #76 does the same thing :facepalm:

It seems like that is pretty stale and might mess with inferSchema for those cases... But does accept a range of null values (not sure how likely it is that a single CSV will have multiple types of null?)

But as is mentioned in that PR and in other things, it really seems like this should be a part of CSV handling as we have large CSV files where it would be too expensive to remap just for remapping nulls.

falaki:
Thanks @addisonj this looks good. I am going to merge it.

addisonj:
Thanks! Currently running from a fork, any ETA on when you will be cutting a new release?

ankurmitujjain:
Is this part of release?

ankurmitujjain:
Not working on Databricks notebook.. I tried to load csv data with ? as null values.

rjurney:
So this seems to still not be working in 2.0.0?

PR #227, opened 2016-01-06
Test against Spark 1.6.0

HyukjinKwon:
Since [Spark 1.6.0](http://spark.apache.org/docs/1.6.0/) is released and this still stays as a third-party library, I think it would be better to test against Spark 1.6.0.
Fortunately, this looks testing fine with this version.


falaki:
Thanks for submitting this.

PR #232, opened 2016-01-16
allow HadoopRDDs minPartitions to be specified

koertkuipers:
We had a need for this to get to a desired number of partitions without a shuffle.

Not sure if this is useful for others as well.


koertkuipers:
Oh just noticed someone already has similar pullreq. Never mind this.

Issue #216, opened 2015-12-23
Incorrect type inference


tanwanirahul:
Assume the following dataset:

A,B,C,D
1,,,
,1,,
,,1,
,,,1

All the columns ideally should be inferred as Integer/Boolean types. But when we use the spark-csv, here is what we get:

root
 |-- A: integer (nullable = true)
 |-- B: integer (nullable = true)
 |-- C: string (nullable = true)
 |-- D: string (nullable = true)

Columns C & D are inferred as String type. Is this really the expected behavior? Any justification would be helpful.

falaki:
I see space before and after third delimiter. That could be the reason.

tanwanirahul:
If you parse the same csv with SparkContext's textFile, you would get the length of the first records as 7, whilst the rest of them have the length of 4. This indicates there isn't any white space?

tanwanirahul:
Here is another dataset example:

A,B,C,D
1,,1,

and the inferred schema:
root
 |-- A: string (nullable = true)
 |-- B: string (nullable = true)
 |-- C: string (nullable = true)
 |-- D: string (nullable = true)

A & C should be Integers here.

tanwanirahul:
@falaki Could you please check on the use cases shared above and confirm if there is something that needs fix? If there is an issue, I feel it must be resolved ASAP because its very fundamental.

falaki:
Yes, this is a bug. We will hopefully publish a new release with a bug fix. You are more than welcome to submit a fix. The bug is in `InferSchema.scala`

tanwanirahul:
Cool. Have plenty on my plate for the week. Will try to submit by early/mid next week.

AEDWIP:
Thanks Gentlemen. 

Happy New Year

Andy

From:  Rahul <notifications@github.com>
Reply-To:  databricks/spark-csv
<reply+00725a2ffd7390630b0e8afd240275f4cee4ac4dfa2019c592cf00000001129a5fab9
2a169ce075eaca0@reply.github.com>
Date:  Tuesday, December 29, 2015 at 6:50 AM
To:  databricks/spark-csv <spark-csv@noreply.github.com>
Subject:  Re: [spark-csv] Incorrect type inference (#216)

tanwanirahul:
When I use the master branch, the schema inference is working as expected. May be the issue is already fixed. If you remember the pull request that fixed this, would be good to put it into comments. 

For the dataset I shared as a test data:
A,B,C,D
1,,,
[...]

PR #235, opened 2016-01-20
Support to specify a shorten name for compression codec

HyukjinKwon:
#234
This PR is similar with https://github.com/apache/spark/pull/10805. 

This PR adds the support for shorten names for compression codecs and added a `CompressionCodecs` class instead of the implicit function as its use is nor recommended.

PR #244, opened 2016-01-29
Fix schema/type inference issue #216

tanwanirahul:
The issue with current implementation is that, it bumps up the NullType to StringType in reduce phase of each partition. When the final reduce across partitions takes place the column type is inferred as StringType even though the column has only one integer value because of incorrect bumping up of NullTypes.

Here, in this pull, we don't bump up the Null to String as a part of reduce operation (within and across partitions). After the reduction, when we create Fields of StructType, we convert NullTypes to StringTypes.


falaki:
Thanks a lot for submitting this. The code looks good. Would you add another unit test that verify it works end-to-end. Maybe use the input you provided in the issue:
```
A,B,C,D
1,,,
,1,,
,,1,
,,,1
```

tanwanirahul:
Added an end to end test case to validate the proper schema/type inference. @falaki Take a look and merge if everything is fine.

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
```diff
@@ -52,4 +83,14 @@ class InferSchemaSuite extends FunSuite {
       Array(LongType)).deep == Array(DoubleType).deep)
   }
 
+  test("Type/Schema inference works as expected for the simple parse dataset.")
```
HyukjinKwon: Hm.. isn't this a rather end-to-end test?

src/main/scala/com/databricks/spark/csv/CsvParser.scala
```diff
@@ -117,12 +116,12 @@ class CsvParser extends Serializable {
     this
   }
 
-  /** Returns a Schema RDD for the given CSV path. */
```
HyukjinKwon: This is just my personal thought. This refactoring looks pretty reasonable in a way though.. Could we maybe do this in a separate PR?

tanwanirahul: Okay. Would make more sense.

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
tanwanirahul: I agree that CsvSuite performs all end to end tests but since there was a dedicated suite for SchemaInference, I did prefer to put the schema tests in there. Do you see any issues in that?

HyukjinKwon: No, I think it is okay though. I just said this because I see it has been added in this way.

For example, https://github.com/databricks/spark-csv/pull/235 and https://github.com/databricks/spark-csv/pull/224

tanwanirahul: Reverted the refactoring part. Tests I kept as is. If more people feel CsvSuite would rather be the right place to put these tests in, I will make that change.

tanwanirahul:
@HyukjinKwon Created https://github.com/databricks/spark-csv/pull/248 for the refactoring part.

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
```diff
@@ -40,6 +63,14 @@ class InferSchemaSuite extends FunSuite {
     assert(InferSchema.inferField(LongType, "2015-08 14:49:00") == StringType)
   }
 
+  test("Merging Nulltypes should yeild Nulltype.")
+  {
+      assert(
```
falaki: Nit: the indent is off:
```scala
assert(
  InferSchema.mergeRowTypes(Array(NullType),
  Array(NullType)).deep == Array(NullType).deep)
```

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
```diff
@@ -52,4 +83,14 @@ class InferSchemaSuite extends FunSuite {
       Array(LongType)).deep == Array(DoubleType).deep)
   }
 
+  test("Type/Schema inference works as expected for the simple sparse dataset.")
+  {
+    val df = new CsvParser().withUseHeader(true).withInferSchema(true)
+            .csvFile(sqlContext, simpleDatasetFile)
+    assert(
+        df.schema.fields.map{field => field.dataType}.deep ==
```
falaki: Nit: use two spaces for indent. of this line. Next line's indent is fine.

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
```diff
@@ -52,4 +83,14 @@ class InferSchemaSuite extends FunSuite {
       Array(LongType)).deep == Array(DoubleType).deep)
   }
 
+  test("Type/Schema inference works as expected for the simple sparse dataset.")
```
falaki: Please move this test to `CsvSuite.scala`. In that suite, there is already a test SQL context, and you don't need to construct SQLContext in this file.

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
```diff
@@ -2,8 +2,31 @@ package com.databricks.spark.csv.util
...
+import com.databricks.spark.csv.CsvRelation
 
-class InferSchemaSuite extends FunSuite {
+class InferSchemaSuite extends FunSuite with BeforeAndAfterAll {
+
+  private val simpleDatasetFile = "src/test/resources/simple.csv"
+  private val utf8Charset = "utf-8"
+  private var sqlContext: SQLContext = _
+
+  override def beforeAll(): Unit =
```
falaki: These can be removed once you move your end-to-end test to `CsvSuite.scala`. This suite a unit test and does not need to create a SQLContext.

falaki:
@tanwanirahul left some more comments. Once  you address I am going to merge this. After that would you mind submitting a patch to spark. Spark-csv is now inlined in Spark 2.0.

tanwanirahul:
@HyukjinKwon Moved end to end tests to CSVSuite, @falaki Also recommended the same.

@falaki Take a look at the CSVSuite for test case and InferSchemaSuite for Indent comments you left. Also let me know your thoughts on https://github.com/databricks/spark-csv/pull/248 

I will push the patch to spark repo soon.

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -717,6 +717,16 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
 
     assert(results.size === numCars)
   }
+
+  test("Type/Schema inference works as expected for the simple sparse dataset.")
+  {
```
HyukjinKwon: I believe this `{` should be move up to the previous line.

src/test/scala/com/databricks/spark/csv/util/InferSchemaSuite.scala
```diff
@@ -55,6 +55,13 @@ class InferSchemaSuite extends FunSuite {
     assert(InferSchema.inferField(LongType, "2015-08 14:49:00") == StringType)
   }
 
+  test("Merging Nulltypes should yeild Nulltype.")
+  {
```
HyukjinKwon: Ditto.

tanwanirahul:
@HyukjinKwon @falaki Is there any eclipse formatting profile that I could load and apply it for the code I am writing?

tanwanirahul:
@falaki Could we re-build the commit? I don't think build failed because of the code issue.

tanwanirahul:
@falaki @HyukjinKwon Could you please look at what is causing the build to fail? Looking at the build error, I don't think its code issue.

tanwanirahul:
@falaki Could you please suggest next steps for this? Also, is there any date planned for the next release? We are currently using this change by adding an unmanaged dependency.

falaki:
Would you please try that combination of spark and openjdk version to investigate what goes wrong?

HyukjinKwon:
Let me take a look when I have some time.

HyukjinKwon:
Hm.. It looks like this was failed due to the cached jars in travis.ci. This works fine both on my local and travis with a new build [here](https://travis-ci.org/HyukjinKwon/spark-csv/builds/108464686).

I will create a new PR for this. Please make the author @tanwanirahul  when you merge that PR by `merge_pr.py`.

HyukjinKwon:
Can we maybe close this?

Issue #231, opened 2016-01-15
Encoding not working with non-ascii compatible encodings (UTF-16/32 etc.)


HyukjinKwon:
It looks this library has the support for other encodings. However, this library uses Hadoop's [`TextInputFormat`](https://github.com/apache/hadoop/blob/master/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/TextInputFormat.java), which uses [`LineRecordReader`](https://github.com/apache/hadoop/blob/master/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/LineRecordReader.java).

According to [MAPREDUCE-232](https://issues.apache.org/jira/browse/MAPREDUCE-232), it looks [`TextInputFormat`](https://github.com/apache/hadoop/blob/master/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/TextInputFormat.java) does not guarantee all encoding types but officially only UTF-8 (as commented in [`LineRecordReader#L147`](https://github.com/apache/hadoop/blob/master/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/LineRecordReader.java#L147)).

According to [MAPREDUCE-232#comment-13183601] (https://issues.apache.org/jira/browse/MAPREDUCE-232?focusedCommentId=13183601&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13183601), it still looks fine with most encodings though but without UTF-16/32.

I think these better be documented in `README.md`.

HyukjinKwon:
In more details, 

I tested this in Max OS. I converted `cars_iso-8859-1.csv` into `cars_utf-16.csv` as below:

```bash
iconv -f iso-8859-1 -t utf-16 < cars_iso-8859-1.csv > cars_utf-16.csv
```

and run the codes below:

```scala
val cars = "src/test/resources/cars_utf-16.csv"
sqlContext.csvFile(cars, parserLib = parserLib, charset = "utf-16", delimiter = 'þ').show()
```

This produces a wrong results below:
```
+----+-----+-----+--------------------+------+
[...]

PR #246, opened 2016-02-01
Validation for supported encodings (except non-ascii compatible encodings) and some stylistic corrections. 

HyukjinKwon:
#231 

As this library does not support some non-ascii compatible encodings, this is described in `README.md`.

Also, `InferSchema` has the function `infer()` instead of `apply()` as it does not return the corresponding object.
This is not recommended according to [databricks/scala-style-guide/README.md#apply_method] (https://github.com/databricks/scala-style-guide/blob/master/README.md#apply_method)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,20 +43,20 @@ This package allows reading CSV files in local or distributed filesystem as [Spa
 When reading files the API accepts several options:
 * `path`: location of files. Similar to Spark can accept standard Hadoop globbing expressions.
 * `header`: when set to true the first line of files will be used to name columns and will not be included in data. All types will be assumed string. Default value is false.
-* `delimiter`: by default columns are delimited using `,`, but delimiter can be set to any character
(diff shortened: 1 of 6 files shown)


HyukjinKwon:
I thought there are two options.
1. Just describe that this Datasource does not support non-ascii compatible encodings (UTF-16/32 etc.)
2. Try to convert \n and compare with \n in UTF-8 in order to validate the encodings.
But, I choose the first one.

HyukjinKwon:
Actually I think both would be great. I added a commit for validating the encoding types.

HyukjinKwon:
Please do not merge this until [SPARK-13108](https://issues.apache.org/jira/browse/SPARK-13108) is resolved.

HyukjinKwon:
Closed for now as we do not have a proper conclusion to fix this yet.

PR #250, opened 2016-02-02
Optimizations for a variable and size() for length().

HyukjinKwon:
I forgot to correct the variable in a past PR. This PR re-use the `rowArray` rather than creating it everytime. This was mentioned before in https://github.com/databricks/spark-csv/pull/169#discussion_r42461113.

In addition, according to IntelliJ IDE, we might have to use `length()` instead of `size()`.
```
This inspection reports array.size and string.size calls.

While such calls are legimimate, they require an additional implicit conversion to SeqLike to be made.

Considering the common use cases, calling length on arrays and strings may provide significant advantages.
```

Lastly, I gave a `()` for `buildScan()`.


HyukjinKwon:
I see this was also fixed in Spark, https://github.com/apache/spark/commit/e97fc7f176f8bf501c9b3afd8410014e3b0e1602.

@falaki Would you maybe merge this?

PR #252, opened 2016-02-02
NullPoingException in schema inference for CSV when the first line is empty

HyukjinKwon:
https://issues.apache.org/jira/browse/SPARK-13137

This PR adds a filter in schema inference so that it does not emit NullPointException.

Also, I removed `MAX_COMMENT_LINES_IN_HEADER `but instead used a monad chaining with `filter()` and `first()`.

Lastly, I simply added a newline rather than adding a new file for this so that this is covered with the original tests.


HyukjinKwon:
@falaki Could you review this please? I believe newlines are being ignored in this library and like this way, we can just skip the first empty newlines.

Also, I think we can unset `MAX_COMMENT_LINES_IN_HEADER`. This would not harm performance.

falaki:
Thanks this looks good.

PR #254, opened 2016-02-03
added support for saving with a QuoteMode

### LINKED ISSUE #119: Write CSV without quotes
I am looking to write a CSV file with out quotes. Default is ". When I tried it with .option("quotes","") I got error message "only one quote is allowed". 

        If this(writing with out quotes) is not supported in current version any pointers who to implement this would be appreciated.

tobithiel:
I'm dealing with some messy csv files and being able to just quote all fields is very useful, so that other applications don't misunderstand the file because of some sketchy characters.
This also enables disabling quoting all together, which would resolve #119, and only quoting non-numeric fields.

### DIFF
--- a/src/test/scala/com/databricks/spark/csv/CsvSuite.scala
+++ b/src/test/scala/com/databricks/spark/csv/CsvSuite.scala
@@ -442,6 +442,36 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
     assert(carsCopy.collect.map(_.toString).toSet == cars.collect.map(_.toString).toSet)
   }
 
+  test("DSL save with a quoteMode") {
+    // Create temp directory
+    TestUtils.deleteRecursively(new File(tempEmptyDir))
+    new File(tempEmptyDir).mkdirs()
+    val copyFilePath = tempEmptyDir + "cars-copy.csv"
+
+    val cars = sqlContext.csvFile(carsFile, parserLib = parserLib)
+    cars.saveAsCsvFile(copyFilePath, Map("header" -> "true", "quoteMode" -> "ALL"))
+
+    val carsCopy = sqlContext.csvFile(copyFilePath + "/")
+
+    assert(carsCopy.count == cars.count)
+    assert(carsCopy.collect.map(_.toString).toSet == cars.collect.map(_.toString).toSet)
+  }
+
+  test("DSL save with non numeric quoteMode") {
+    // Create temp directory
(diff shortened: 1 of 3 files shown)


src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -442,6 +442,51 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
...
   }
 
+  test("DSL save with a quoteMode") {
+    // Create temp directory
+    TestUtils.deleteRecursively(new File(tempEmptyDir))
+    new File(tempEmptyDir).mkdirs()
+    val copyFilePath = tempEmptyDir + "cars-copy.csv"
+
+    val cars = sqlContext.csvFile(carsFile, parserLib = parserLib)
+    cars.saveAsCsvFile(copyFilePath, Map("header" -> "true", "quoteMode" -> "ALL"))
```
HyukjinKwon: Hm.. Does this really check if the output are quoted or not?

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -442,6 +442,51 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
...
+
+    val carsCopy = sqlContext.csvFile(copyFilePath + "/")
+
+    assert(carsCopy.count == cars.count)
+    assert(carsCopy.collect.map(_.toString).toSet == cars.collect.map(_.toString).toSet)
```
HyukjinKwon: Hm.. Does this really check if the output are quoted or not?

tobithiel:
Not sure why it's not rerunning the tests, but I pushed extended tests. I'm not that familiar with Scala, so please feel free to write that in a more elegant way.
Most of the other tests writing csv files don't check what's actually in the file as well, so these kind of checks should probably extended to these tests as well.

tobithiel:
@HyukjinKwon Are you interested in merging this in? Do you have more feedback?

HyukjinKwon:
@tobithiel I do not have the permission. cc @falaki

falaki:
@tobithiel This is a cool feature to have in spark-csv. Would you make sure it works with the other parse mode as well?

tobithiel:
@falaki I'm not sure I get what you mean? This only deals with writing files and since parsing csv files with quotes is already tested & works, parsing the written files again works out of the box.

falaki:
@tobithiel Thanks. This looks good.

Issue #141, opened 2015-08-31
Number of partitions.


stati:
Is it possible to set no.of partitions that will be created when reading a CSV?
Similar to minPartitions in sc.textFile("/Path/to/file",40)

PR #256, opened 2016-02-03
support for passing minPartitions when reading file

tobithiel:
It's quite expensive to have to reshuffle huge files after reading to partition them better, so this allows passing the minPartitions argument Hadoop/Spark provide to automatically partition the file as needed while reading.
#141 and #186 already request/address this. You don't seem to be in favor of adding this, but without exposing this option (already present in Spark/Hadoop), this can't be done with the same efficiency, I believe.

The other pull request contains a lot of unrelated/-necessary changes, so I'm hoping to reopen this discussion with a cleaner diff.
Let me know what you think.


tobithiel:
Had to fix some things my local tests didn't show, I'll squash these together if you want to merge it.

HyukjinKwon:
I opened a Jira issue for Spark, [here](https://issues.apache.org/jira/browse/SPARK-13184) for discussion.

HyukjinKwon:
Can we maybe close this for now until we have a proper conclusion in the JIRA I made?

PR #258, opened 2016-02-09
report the error as well as the line that caused it

abridgett:
I don't know Scala (at all!) so there's almost certainly cleaner ways - my apologies.  The logging at the moment is sometimes unhelpful as it's hard to see the real issue - with DROPMALFORMED you see the line, with another parsing mode you get the error but not the line.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -105,7 +105,7 @@ case class CsvRelation protected[spark] (
     tokenRdd(schemaFields.map(_.name)).flatMap { tokens =>
 
       if (dropMalformed && schemaFields.length != tokens.size) {
-        logger.warn(s"Dropping malformed line: ${tokens.mkString(",")}")
+        logger.warn(s"Dropping malformed line (wrong length): ${tokens.mkString(",")}")
         None
       } else if (failFast && schemaFields.length != tokens.size) {
         throw new RuntimeException(s"Malformed line in FAILFAST mode: ${tokens.mkString(",")}")
@@ -125,13 +125,16 @@ case class CsvRelation protected[spark] (
           case aiob: ArrayIndexOutOfBoundsException if permissive =>
             (index until schemaFields.length).foreach(ind => rowArray(ind) = null)
             Some(Row.fromSeq(rowArray))
-          case _: java.lang.NumberFormatException |
-               _: IllegalArgumentException if dropMalformed =>
-            logger.warn("Number format exception. " +
+          case e: java.lang.NumberFormatException if dropMalformed =>
+            logger.warn("Number format exception  (" + e + ")." +
+              s"Dropping malformed line: ${tokens.mkString(delimiter.toString)}")
+            None
+          case e: IllegalArgumentException if dropMalformed =>
+            logger.warn("IllegalArgument exception  (" + e + ")." +
               s"Dropping malformed line: ${tokens.mkString(delimiter.toString)}")
             None
           case pe: java.text.ParseException if dropMalformed =>
-            logger.warn("Parse exception. " +
+            logger.warn("Parse exception. (" + pe + ")." +
               s"Dropping malformed line: ${tokens.mkString(delimiter.toString)}")
             None
         }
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -170,7 +173,7 @@ case class CsvRelation protected[spark] (
       val requiredSize = requiredFields.length
       tokenRdd(schemaFields.map(_.name)).flatMap { tokens =>
         if (dropMalformed && schemaFields.length != tokens.size) {
-          logger.warn(s"Dropping malformed line: ${tokens.mkString(delimiter.toString)}")
+          logger.warn(s"Dropping malformed line (wrong length): ${tokens.mkString(delimiter.toString)}")
```
HyukjinKwon: Please check travis. `[error] /home/travis/build/databricks/spark-csv/src/main/scala/com/databricks/spark/csv/CsvRelation.scala:176: File line length exceeds 100 characters`

HyukjinKwon:
For me I think this is reasonable because I sometimes see some users have some difficulties to find out why it was dropped or malformed.

abridgett:
Thanks Hyukjin, I've fixed that overlong line

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -125,13 +125,16 @@ case class CsvRelation protected[spark] (
...
+            logger.warn("Number format exception  (" + e + ")." +
+              s"Dropping malformed line: ${tokens.mkString(delimiter.toString)}")
+            None
+          case e: IllegalArgumentException if dropMalformed =>
+            logger.warn("IllegalArgument exception  (" + e + ")." +
               s"Dropping malformed line: ${tokens.mkString(delimiter.toString)}")
             None
           case pe: java.text.ParseException if dropMalformed =>
-            logger.warn("Parse exception. " +
+            logger.warn("Parse exception. (" + pe + ")." +
```
HyukjinKwon: Personal opinion: since this is a warning message, shouldn't we maybe just print out the message only by `pe.getMessage()`?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -125,13 +125,16 @@ case class CsvRelation protected[spark] (
           case aiob: ArrayIndexOutOfBoundsException if permissive =>
             (index until schemaFields.length).foreach(ind => rowArray(ind) = null)
             Some(Row.fromSeq(rowArray))
-          case _: java.lang.NumberFormatException |
-               _: IllegalArgumentException if dropMalformed =>
-            logger.warn("Number format exception. " +
+          case e: java.lang.NumberFormatException if dropMalformed =>
+            logger.warn("Number format exception  (" + e + ")." +
```
HyukjinKwon: Personal opinion: since this is a warning message, shouldn't we maybe just print out the message only by `pe.getMessage()`?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -105,7 +105,7 @@ case class CsvRelation protected[spark] (
     tokenRdd(schemaFields.map(_.name)).flatMap { tokens =>
 
       if (dropMalformed && schemaFields.length != tokens.size) {
-        logger.warn(s"Dropping malformed line: ${tokens.mkString(",")}")
+        logger.warn(s"Dropping malformed line (wrong length): ${tokens.mkString(",")}")
```
HyukjinKwon: Personal opinion: likewise, shouldn't we maybe need a more detail message rather than just printing `wrong length`? e.g.` Expected token length is 5 but the parsed token length was 4`.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
HyukjinKwon: In addition, you can write `s"Number format exception  (${a.getMessage})."` instead of concatenating with `+` which is not used generally in this library and Spark codes as well.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
abridgett: Sorry for the delay - I've fixed those other messages, but here is it not useful to know which line has caused the error.  Compared to the errors below it's probably easier to find this in the original files, however it's still even easier if we tell the user about the problems lines :-)

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -205,7 +205,7 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
         .collect()
     }
 
-    assert(exception.getMessage.contains("Malformed line in FAILFAST mode: 2015,Chevy,Volt"))
+    assert(exception.getMessage.contains("Malformed line in FAILFAST mode (expected 5 tokens but received 3 tokens): 2015,Chevy,Volt"))
```
falaki: This line exceeds 100 characters as well.

PR #259, opened 2016-02-10
Change a few access modifiers to improve extensibility.

blrnw3:
Currently a lot of functionality is hidden behind private access modifiers making it difficult to extend the functionality of the library without copying large chunks of code. These small changes make re-use much easier.

The specific use case I have is parsing fixed-width text files. The univosity parser library (used by spark-csv) actually supports this, but spark-csv itself does not. Rather than rewriting a good chunk of the existing library I opted to extend it by adding a new entry point and swapping-out the parsers. To do this, I had to make a few methods and classes in spark-csv more accessible. Hopefully this also makes extensibility in general easier, so parsers for other file types can be more easily created. 

Also, If there's general interest in flat-file parsing then I'm happy to incorporate that directly into spark-csv.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -28,7 +28,7 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
 import org.apache.spark.sql.sources.{PrunedScan, BaseRelation, InsertableRelation, TableScan}
 import org.apache.spark.sql.types._
-import com.databricks.spark.csv.readers.{BulkCsvReader, LineCsvReader}
+import com.databricks.spark.csv.readers.{BulkCsvReader, LineCsvReader, BulkReader, LineReader}
 import com.databricks.spark.csv.util._
 
 case class CsvRelation protected[spark] (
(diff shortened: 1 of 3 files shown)


pheuter:
:+1:

HyukjinKwon:
For me, I thinks it is questionable. I think we can't expose some classes or functionalities only because the functionalities are needed by some users.

This is my personal opinion though. Shouldn't we maybe create an issue to discuss which functionalities should be exposed or not? There might be some reasons why they are hidden.

blrnw3:
Perhaps spark-csv could become spark-flatfile, and add support for fixed-width parsing. I thought about doing this but it involved a significant refactoring of the codebase which I did not think would be admissible. Hence I went for the option of making small changes so I could spin-off from spark-csv and make a spark-fixedwidth project (https://github.com/quartethealth/spark-fixedwidth/tree/b9e60e4e6473c14c78f0b11ff8e5c7514c150959). 

However, if you think adding fixed-width support is a good idea, I am happy to work on that and integrate it into spark-csv.

falaki:
If we expose the new API in spark-csv we need to support them in all future versions and we cannot remove them. That is why we are very careful about exposing new API. In fact, at this point I wish I had not exposed the CsvParser API.

HyukjinKwon:
Can we maybe close this for now?

PR #261, opened 2016-02-11
Fix schema/type inference issue

HyukjinKwon:
https://github.com/databricks/spark-csv/pull/244
This is re-opened due to the build failure in travis.


tanwanirahul:
@HyukjinKwon @falaki Do you want me to raise new PR with a single commit? That will make sure we have a clean commit history?

HyukjinKwon:
@tanwanirahul `dev/merge_pr.py` will make this as a single commit log as PRs are not merged by simply pushing the merge button and these commit logs will only stay in comments in a single commit log at the end with a single author.

This is why I said here https://github.com/databricks/spark-csv/pull/244#issuecomment-182734399, the author might have to be you.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -42,7 +42,11 @@ private[csv] object InferSchema {
       mergeRowTypes)
 
     val structFields = header.zip(rootTypes).map { case (thisHeader, rootType) =>
-      StructField(thisHeader, rootType, nullable = true)
+      val dType = rootType match {
```
HyukjinKwon: BTW, does this not produce `Nulltype` after merging `Nulltype`s?

HyukjinKwon:
Well, but If you need to change some codes here, then I think you better create another PR maybe.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
tanwanirahul: @HyukjinKwon Please elaborate. Referring to doc comments in InferSchema.scala

  /**
   * Similar to the JSON schema inference.
   * [[org.apache.spark.sql.execution.datasources.json.InferSchema]]
   *     1. Infer type of each row
   *     2. Merge row types to find common type
   *     3. Replace any null types with string type
   */
All null types should be replaced with String types at the end. This is what is happening right now?
If I understand you correctly, you feel there is a need for some more tests to be added. Could you please mention what that test is suppose to test?

As far as testing merging two nulltypes returning null type, is already covered.

HyukjinKwon: Ah, sorry. Please ignore this comment.

tanwanirahul:
I will creata a JIRA on Spark and submit a patch once this is merged.

falaki:
Thanks. This looks good and passes all tests. Merging it now.

PR #263, opened 2016-02-12
Update library name to "CSV Data Source for Spark"

mengxr:
Calling the library "Spark CSV" might confuse users about the source of the library, while doesn't conform to ASF trademark policy. This PR renames it to "CSV Data Source for Spark".


falaki:
Thanks.

Issue #188, opened 2015-11-13
support for transforming dataframe to RDD[String]


pjfanning:
Thanks for all the hard work on spark-csv. It is a very useful library, as CSV is a really nice format to get up and running with Spark dataframes.
https://github.com/databricks/spark-csv/commit/2eb90153a2d6a77b9cde4aee3f6e382df3da1746 adds a csvRdd function to create a Dataframe from RDD[String]. It would be useful for me to be able to the reverse.
https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/package.scala saveAsCsvFile already has this but the strRdd is just a temporary internal state used to save the csv to file.
I can put together a pull request with a new function saveAsCsvRdd or convertToCsvRdd if that sounds like an reasonable idea.
Also, is there a plan to publish a new 1.3 release in the near future?

pjfanning:
I have a sample commit at https://github.com/pjfanning/spark-csv/commit/376568acb7f975fe69c26ae790c6792b0d4d002c - I can add more tests if this change is going down the right track

falaki:
Thanks for your suggestion. I just published version 1.3.0.

Tagar:
It would be great to have this built into Spark-CSV or into Spark. Thank you @HyukjinKwon

pjfanning:
I created a pull request: https://github.com/databricks/spark-csv/pull/270

Issue #264, opened 2016-02-14
infer schema using spark-csv over parquet table with string columns


Tagar:
Would be great to use spark-csv to infer schema from an arbitrary DF or a parquet table that has string columns. Or is this already possible perhaps?

HyukjinKwon:
For me, I am not too sure if we need this but It is anyway possible. 
First of all, you might be able to do this by converting data, `DataFrame` -> `RDD[String]` -> `DataFrame` as below (if this has nested data types, this will produces wrong data):

```scala
val rdd = arbitraryDf.rdd.map(_.toSeq.mkString(","))
val df = new CsvParser().csvRdd(sqlContext, rdd)
```

Tagar:
> map(_.toSeq.mkString(","))
wouldn't really work because some of the strings might have commas.
have a look at http://stackoverflow.com/questions/32722132/spark-data-type-guesser-udaf
thank you.

HyukjinKwon:
@Tagar I meant to say this workaround is anyway possible in this way.
For example, we can escape from the commnas with this.
```scala
val rdd = arbitraryDf.rdd.map(_.toSeq.map(v => s""""$v"""").mkString(","))
val df = new CsvParser().csvRdd(sqlContext, rdd)
```
If this contains `"` then, we can set another separator for this and use `quote` option for reading this.
If converting `DataFrame` to `RDD[String]` can be done (filed here https://github.com/databricks/spark-csv/issues/188), we might be able to do this.

Tagar:
I have to agree. This is possible, but looks messy and requires two passes / two conversions:
convert dataframe to... csv just to make it work, and convert that back to a dataframe.

What I am suggesting in this issue requires one pass.
We have a lot of use cases when very wide datasets have ~98% of columns being just strings,
because source systems don't give us schema.

It would be great to have a feature of schema inference over RDD of strings built into 
Spark-CSV or into Spark. Thank you.

Tagar:
Just noticed #188 thank you @HyukjinKwon

PR #270, opened 2016-02-22
Rdd string

pjfanning:
Relates to https://github.com/databricks/spark-csv/issues/188


falaki:
The implicit functions in `package.scala` are not going to be updated in future. Users are encouraged to use spark's `load` and `save` API to interact with spark-csv rather than directly calling the implicit functions.

Tagar:
Thank you. Which spark-csv version this PR will be part of?

falaki:
load and save functions exist in all Spark 1.4+ versions.

Tagar:
I was asking regarding "adds a csvRdd function to create a Dataframe from RDD[String]".
Please check https://github.com/databricks/spark-csv/issues/264
Thanks.

PR #271, opened 2016-02-23
Fixed setting escape character with none quoteMode.

d18s:
Swapped the order of calls to withQuoteMode(quoteMode) and withEscape(escapeChar)
to make sure that the escape character is set before setting the quoteMode, which
requires an escape character to be set to be valid.

### DIFF
--- a/src/test/scala/com/databricks/spark/csv/CsvSuite.scala
+++ b/src/test/scala/com/databricks/spark/csv/CsvSuite.scala
@@ -26,6 +26,7 @@ import org.apache.spark.sql.{SQLContext, Row, SaveMode}
 import org.apache.spark.{SparkContext, SparkException}
 import org.apache.spark.sql.types._
 import org.scalatest.{BeforeAndAfterAll, FunSuite}
+import org.scalatest.Matchers._
 
 abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
   val carsFile = "src/test/resources/cars.csv"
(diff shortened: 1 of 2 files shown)


falaki:
Thanks for the patch. Can you think of a case that the old order did not function properly. Would you add a unit test to your patch for such a case?

d18s:
Hi falalki, sure thing, will add a unit test.

If you set `quoteMode="None"` it will always fail if `withQuoteMode(quoteMode)` is called before `withEscape(escapeChar)` and the default `escapeChar` is `null` due to this check in the `validate()` function in Apache commons CSVFormat.java:

```
if (escapeCharacter == null && quoteMode == QuoteMode.NONE) {
            throw new IllegalArgumentException("No quotes mode set but no escape character is set");
}
```

i.e., you have to set the `escapeChar` (to a non null value) before the `quoteMode`.

d18s:
Hi falaki, I've added a test and squashed the fix and the test commits into one.

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -547,6 +548,24 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
...
+    new File(tempEmptyDir).mkdirs()
+    val copyFilePath = tempEmptyDir + "cars-copy.csv"
+
+    val cars = sqlContext.csvFile(carsFile, parserLib = parserLib)
+    val delimiter = ","
+    var quote = "\""
+    var escape = "\\"
+    var quoteMode = "NONE"
+
+    noException should be thrownBy {
```
HyukjinKwon: I know those things such as `should be` are pretty much used in Scala as some kind of convention. However, would this be better if we have some test codes having the same format?

I have seen such things used in various open sources but not for Spark yet.

falaki:
Thanks. This looks goods and closes https://github.com/databricks/spark-csv/issues/278.

Issue #279, opened 2016-03-01
need option for specifying date format 


barrybecker4:
I noticed these comments in the code:
// TODO(hossein): would be good to support other common timestamp formats
case _: TimestampType => Timestamp.valueOf(datum)
// TODO(hossein): would be good to support other common date formats
case _: DateType => Date.valueOf(datum)

It would be really nice to be able to specify the format of date columns when reading in a CSV file.
For example, something like this:

df = sqlContext.read
                .format("com.databricks.spark.csv")
                .option("inferSchema", "false") 
                .option("dateFormat", "mm/dd/yyyy hh:mm:ss")
                .schema(dfSchema)
                .load(dataFile)

Currently the only format allowed is (yyyy-mm-dd hh:m:ss")
Is there a workaround for this besides just changing the data before hand? I suppose you could read all date columns as string then operate on the dataframe to change them all to dates, but it would be pretty tedious.

HyukjinKwon:
I think this is a duplicate of https://github.com/databricks/spark-csv/issues/262?

barrybecker4:
Yes, it's basically the same as #262. Sorry I did not see it. One can be duped to the other.

HyukjinKwon:
@barrybecker4 Could we maybe close this? This is just fixed in #280.

barrybecker4:
closing.

barrybecker4:
I realize that this issue has been closed, but I had two questions.
1) Assuming that the reading and writing of a file is happening in parallel, I think it would be safer to use Joda's DateTimeFormat (which is threadsafe) instead SimpleDateFormat (which is not threadsafe). Is this a concern?
2) Is the dateFormat used when writing? It appears that it is not. I will open a separate issue for this.
Here is what I am doing and it does not write in the specified format:
repartitionDf.write.format("com.databricks.spark.csv")
        .option("delimiter", "\t")
        .option("header", "false")
        .option("nullValue", "?")    // this worked
        .option("dateFormat", "MM/dd/yyyy HH:mm:ss")
        .save(tempFileName)
The file has dates like "2012-01-03 09:12:00.0" but it should be 03/01/2012 09:12:00"

Issue #266, opened 2016-02-16
Inferschema


saritseal:
The following are the issues with Infer schema:
1. It will help if the class can be passed as the parameter and spark-csv uses the custom class to Infer the schema.
2. There is no support for Date, DateTime with a custom format
3. Number formats should have a option to specify the thousand separator, as well as negative formats

m-stafford:
I've recently added some support for basic Date formats to my fork of spark-csv. Tests are still needed before I'd be comfortable submitting a pull request but is this something that would be useful in the master branch?

falaki:
Yes, please feel free to submit your PR.

saritseal:
No worries. We are not using spark-csv I have written a schema inferred
which generates a parquet schema. I am sotastically identifying the sample
based on history and using to identify the schema

HyukjinKwon:
@m-stafford @falaki Could I try if you guys are not working on this?

falaki:
Sure. I think the most important one for now is Date and time format.

HyukjinKwon:
I will submit a PR within tomorrow.

HyukjinKwon:
@falaki @m-stafford @saritseal  Could we maybe close this? This is just fixed in #280.

Issue #262, opened 2016-02-12
Custom Timestamp format


aplavin:
How can I read CSV files with Timestamps formatted like `1966-02-13T00:00:00` (note `T` as the separator)? I see no way to specify the format, and reading as-is obviously gives `java.lang.IllegalArgumentException: Timestamp format must be yyyy-mm-dd hh:mm:ss[.fffffffff]`.

Thanks!

HyukjinKwon:
I think this is the subset of https://github.com/databricks/spark-csv/issues/266

HyukjinKwon:
@aplavin  Could we maybe close this? This is just fixed in #280.

raam86:
This was not fixed in 280, it's easily replicated in spark 2.1 and spark csv 1.5.0

PR #280, opened 2016-03-04
Support for specifying custom date format for date and timestamp types.

HyukjinKwon:
https://github.com/databricks/spark-csv/issues/279
https://github.com/databricks/spark-csv/issues/262
https://github.com/databricks/spark-csv/issues/266

This PR adds the support to specify custom date format for `DateType` and `TimestampType`.

For `TimestampType`, this uses the given format to infer schema and also to convert the values
For `DateType`, this uses the given format to convert the values.
If the `dateFormat` is not given, then it works with `Timestamp.valueOf()` and `Date.valueOf()` for backwords compatibility.
When it's given, then it uses `SimpleDateFormat` for parsing data.

In addition, `IntegerType`, `DoubleType` and `LongType` have a higher priority than `TimestampType` in type inference. This means even if the given format is `yyyy` or `yyyy.MM`, it will be inferred as `IntegerType` or `DoubleType`. Since it is type inference, I think it is okay to give such precedences.


HyukjinKwon:
@falaki Just to let you know, the original functions, `tryParse...()`  were not mostly modified but just moved into `inferField()` as inner functions at `src/main/scala/com/databricks/spark/csv/util/InferSchema.scala`.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -128,6 +128,8 @@ class DefaultSource
     val charset = parameters.getOrElse("charset", TextFile.DEFAULT_CHARSET.name())
     // TODO validate charset?
 
+    val dataFormat = parameters.getOrElse("charset", TextFile.DEFAULT_CHARSET.name())
```
falaki: This line needs to be removed.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -76,7 +77,61 @@ private[csv] object InferSchema {
    */
   private[csv] def inferField(typeSoFar: DataType,
     field: String,
-    nullValue: String = ""): DataType = {
+    nullValue: String = "",
+    dateFormatter: SimpleDateFormat = null): DataType = {
+    def tryParseInteger(field: String): DataType = if ((allCatch opt field.toInt).isDefined) {
```
falaki: Indent is off for this entire block

falaki:
@HyukjinKwon left one more comment. Otherwise looks good. I can merge this before I cut the branch tonight.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
HyukjinKwon: Um.. Do you mean the indentation correction as below?
```scala
  private[csv] def inferField(typeSoFar: DataType,
    field: String,
    nullValue: String = "",
    dateFormatter: SimpleDateFormat = null): DataType = {
    def tryParseInteger(field: String): DataType = if ((allCatch opt field.toInt).isDefined) {
      IntegerType
    } else {
      tryParseLong(field)
    }
...
```

```scala
  private[csv] def inferField(typeSoFar: DataType,
    field: String,
    nullValue: String = "",
    dateFormatter: SimpleDateFormat = null): DataType = {
      def tryParseInteger(field: String): DataType = if ((allCatch opt field.toInt).isDefined) {
        IntegerType
      } else {
        tryParseLong(field)
      }
...
```

falaki: Oh I see. The problem is for lines above:
```scala
  private[csv] def inferField(typeSoFar: DataType,
      field: String,
      nullValue: String = "",
      dateFormatter: SimpleDateFormat = null): DataType = {
    def tryParseInteger(field: String): DataType = if ((allCatch opt field.toInt).isDefined) {
      IntegerType
    } else {
      tryParseLong(field)
    }
```

HyukjinKwon: I see. Thanks!

HyukjinKwon:
@falaki (although this is not the right thread to say this), what do you think about `uncompressed` and `none` options for compression codecs to set explicitly no compression?

This was merged in Spark, https://github.com/apache/spark/pull/11464 and I was working on this for this library as well. 

However, I just realised that this might not be some kind of what we must do identically for this, like `compression` option is supported for `codec` at Spark CSV data source but `compression` is not acceptable at this library.

falaki:
Let's open an issue for it.

HyukjinKwon:
Filed in https://github.com/databricks/spark-csv/issues/281.

barrybecker4:
Thank you for adding this! I will pull and build a local snapshot until it is 1.4.0 officially releases.

HyukjinKwon:
@barrybecker4 Maybe would you create a PR for that typo (also for other typos if you know)?

5ean:
I still face error in python. Maybe i did not use it correctly. Please kindly advise.
`df = sqlContext.read.format('com.databricks.spark.csv').options(header='true',dateFormat='dd/MM/yyyy hh:mm:ss').load('test.csv',schema = schema)`

schema has timestamp type.

And the string in csv file '"25/02/2014 00:00:00"

exception:

Caused by: java.lang.IllegalArgumentException: Timestamp format must be yyyy-mm-dd hh:mm:ss[.fffffffff]

raam86:
The docs are not clear on how to do this so hopefully this can help:

https://stackoverflow.com/questions/43259485/how-to-load-csvs-with-timestamps-in-custom-format

```
spark.read
      .option("header", true)
      .option("inferSchema", true)
      .option("timestampFormat", "MM/dd/yyyy h:mm:ss a")
      .csv("PATH_TO_CSV")
```

shatestest:
this is not working as expected.  https://stackoverflow.com/questions/55965978/how-to-set-jdbc-partitioncolumn-type-to-date-in-spark-2-4-1/55966481#55966481

Issue #281, opened 2016-03-04
Support for explicitly setting no compression such as none and uncompressed


HyukjinKwon:
`uncompressed` and `none` options for compression codecs might be needed to set explicitly no compression.

This was merged in Spark; however, I just realised that this might not be some kind of what we must do identically for this, like `compression` option is supported for `codec` at Spark CSV data source but `compression` is not acceptable at this library.

For me I think it is good to do.

PR #283, opened 2016-03-04
Support for explicitly setting no compression via option

HyukjinKwon:
#281 
This PR adds the support for compression options, `uncompressed` and `none` to explicitly set no compression codec.

Also, I mistakenly forgot to add `DeflateCodec` before. This PR also adds the compression codec.


HyukjinKwon:
Although I am not too sure if both `uncompressed` and `none` are what this library must need, I still think it is good to have.

HyukjinKwon:
If this is merged (if you think this is needed), then I will also submit a PR to `spark-xml as well`.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -335,9 +335,20 @@ case class CsvRelation protected[spark] (
       }
       // Write the data. We assume that schema isn't changed, and we won't update it.
 
-      val codecClass = CompressionCodecs.getCodecClass(codec)
-      data.saveAsCsvFile(filesystemPath.toString, Map("delimiter" -> delimiter.toString),
-        codecClass)
+      val hadoopConf = data.sqlContext.sparkContext.hadoopConfiguration
+      val maybeCodecClass = CompressionCodecs.getCodecClass(codec)
+      val parameters = Map("delimiter" -> delimiter.toString)
+      maybeCodecClass match {
```
falaki: I think this can be simpler:
```Scala
maybeCodecClass match {
  case Some(codeClass) if codecClass != null => 
    data.saveAsCsvFile(filesystemPath.toString, parameters, codecClass)
  case _ => 
    data.saveAsCsvFile(filesystemPath.toString, parameters)
}
```

src/main/scala/com/databricks/spark/csv/util/CompressionCodecs.scala
```diff
@@ -32,18 +36,35 @@ private[csv] object CompressionCodecs {
...
         case e: ClassNotFoundException =>
           throw new IllegalArgumentException(s"Codec [$codecName] is not " +
             s"available. Known codecs are ${shortCompressionCodecNames.keys.mkString(", ")}.")
       }
   }
+
+  /**
+   * Disable compression codec.
+   */
+  def disableCompressConfiguration(hadoopConf: Configuration): Unit = {
```
falaki: This seems to have broad side-effects. Why do we need to do this?

HyukjinKwon: Oh, sorry. Thanks.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
HyukjinKwon: I meant to explicitly set non-compression with ignoring global configuration so that it can be ignored by setting `compress` to `none` , even if the compression configuration is set to `sqlContext.sparkContext.hadoopConfiguration`.

Sorry, this would change the global configuration. Let me correct this.

HyukjinKwon:
The commits I just submitted include removing the  broad side-effects and some tests for that, and move codec validation into `saveAsCsvFile()` just like other options.

As `saveAsTextFile()` does not accept Hadoop configuration as a parameter, it just resembles the internal logics in `saveAsTextFile()` by calling `saveAsHadoopFile()`.

In `saveAsCsvFile()`, `case` clause deals with the cases below:

- `codec` is set to `uncompressed`/`none` :  In this case, make it ignore the existing compress configuration of Hadoop but write output as uncompressed without side-effects.
- `codec` is set to other compress codecs :  Set compress configuration by `codec`.
- `codec` is not set :
  - `compressionCodec` is given as a parameter: Set compress configuration by `compressionCodec` .
  - `compressionCodec` is not given as a parameter: Do not set anything but follow the existing compress configuration of Hadoop.

HyukjinKwon:
I had to change pretty a lot of codes and it looks a bit messy (partly because this one does not implement `HadoopFsRelation`). 
I feel like maybe, it does not necessarily have to deal with the cases for `none` and `uncompressed` even if this behaviour is not identical with Spark one...

HyukjinKwon:
The commits I just submitted include removing the  broad side-effects and some tests for that, and move codec validation into `saveAsCsvFile()` just like other options.

As `saveAsTextFile()` does not accept Hadoop configuration as a parameter, it just resembles the internal logics in `saveAsTextFile()` by calling `saveAsHadoopFile()`.

In `saveAsCsvFile()`, [`case` clause(L193-L209)](https://github.com/HyukjinKwon/spark-csv/blob/support-uncompressed/src/main/scala/com/databricks/spark/csv/package.scala#L193-L209) below:

```scala
maybeCodecClass match {
  case Some(codecClass) if codecClass == null =>
    // This case means `codec` is set to `uncompressed` or `none`. In this vase, explicitly
    // set the output as uncompressed and ignore existing Hadoop compression configurations.
    CompressionCodecs.disableCompressConfiguration(hadoopConfiguration)
    hadoopPairRDD
      .saveAsHadoopFile(path,
        classOf[NullWritable], classOf[Text], classOf[TextOutputFormat[NullWritable, Text]],
        hadoopConfiguration, None)
  case _ =>
    // If no compression codec is set in `parameters`, then use `compressionCodec` here.
    val safeCodecClass = maybeCodecClass.orElse(Option(compressionCodec))
    hadoopPairRDD
      .saveAsHadoopFile(path,
        classOf[NullWritable], classOf[Text], classOf[TextOutputFormat[NullWritable, Text]],
        hadoopConfiguration, safeCodecClass)
}
```

 deals with the cases below:

- `codec` is set to `uncompressed`/`none` :  In this case, make it ignore the existing compress configuration of Hadoop but write output as uncompressed without side-effects.
- `codec` is set to other compress codecs :  Set compress configuration by `codec`.
- `codec` is not set :
  - `compressionCodec` is given as a parameter: Set compress configuration by `compressionCodec` .
  - `compressionCodec` is not given as a parameter: Do not set anything but follow the existing compress configuration of Hadoop.

HyukjinKwon:
I had to change pretty a lot of codes and it looks a bit messy. 
I feel like maybe it does not necessarily have to deal with the cases for `none` and `uncompressed` even if this behaviour is not identical with Spark one...

HyukjinKwon:
I am closing this because I am not too sure if this change is safe and I feel this is trivial.

Issue #286, opened 2016-03-07
Consider having a NULL_IF_FAIL mode when parsing values


barrybecker4:
Currently there are 3 modes: PERMISSIVE, DROPMALFORMED, and FAIL_FAST.
I use PERMISSIVE, but it still fails if there is a date that does not match the specified date format, or a string in a column that is specified as numeric in the schema. It might be good to have a mode like NULL_IF_FAIL which reads such problematic values as null. Another possibility is allowing multiple date formats so that an alternative can be used if the first fails.

HyukjinKwon:
I also have thought this before altough I felt a bit hesitate to open an issue because there are currently too many options for this library. 
Nevertheless,  maybe `PERMISSIVE` mode should include this. Since this is being failed to read, I personally think it might be okay to add on this behaviour in terms of backward compatibility.

barrybecker4:
Another thought on this. How about being able to specify a "invalidValue" property that is used if parsing was not successful. So instead of throwing a parse exception, the "invalidValue" will be used. That value of the invalidValue may be something like "INVALID" or perhaps the nullValue. If it s the null value, then values that do not parse would be loaded as null. If invalidValue is not set, then behavior would be as it is today.

PR #298, opened 2016-03-19
Adding and option to the csv parser, withParseExecptionAsNull which a…

rachelwarren:
…llows dirty data to be parsed as nulls rather than cause failures. For example, if there is a String in a numeric column, rather than failing it lets that value be null. My experience is that data is often dirty, and some times you want and option that allows mall-formed numbers or dates without causing failures. I have notice a good deal of issues related to this package have to do with similar problems. My implementation is written so that it doesn't affect any of the existing functionality. Instead, it provides an extra option "withParseExceptionAsNull".

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
+++ b/src/main/scala/com/databricks/spark/csv/CsvRelation.scala
@@ -49,7 +49,9 @@ case class CsvRelation protected[spark] (
     inferCsvSchema: Boolean,
     codec: String = null,
     nullValue: String = "",
-    dateFormat: String = null)(@transient val sqlContext: SQLContext)
+    dateFormat: String = null,
+    treatParseExceptionAsNull: Boolean)
+  (@transient val sqlContext: SQLContext)
   extends BaseRelation with TableScan with PrunedScan with InsertableRelation {
 
   // Share date format object as it is expensive to parse date pattern.
@@ -118,7 +120,8 @@ case class CsvRelation protected[spark] (
           while (index < schemaFields.length) {
             val field = schemaFields(index)
             rowArray(index) = TypeCast.castTo(tokens(index), field.dataType, field.nullable,
-              treatEmptyValuesAsNulls, nullValue, simpleDateFormatter)
+              treatEmptyValuesAsNulls, nullValue, simpleDateFormatter,
+              treatParseExceptionAsNull)
             index = index + 1
           }
           Some(Row.fromSeq(rowArray))
@@ -197,7 +200,8 @@ case class CsvRelation protected[spark] (
                 field.nullable,
                 treatEmptyValuesAsNulls,
                 nullValue,
-                simpleDateFormatter
+                simpleDateFormatter,
+                treatParseExceptionAsNull
               )
               subIndex = subIndex + 1
             }
--- a/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/csv/DefaultSource.scala
@@ -125,6 +125,16 @@ class DefaultSource
       throw new Exception("Treat empty values as null flag can be true or false")
     }
 
+    val treatParseExceptionAsNull = parameters.getOrElse(
+      "treatParseExceptionAsNull", "false")
+    val treatParseExceptionAsNullFlag = if (treatParseExceptionAsNull == "false"){
+      false
+    } else if (treatParseExceptionAsNull == "true") {
+      true
+    } else {
+      throw new Exception("Treat parse exception as null flag can be true or false")
+    }
+
     val charset = parameters.getOrElse("charset", TextFile.DEFAULT_CHARSET.name())
     // TODO validate charset?
 
@@ -159,7 +169,8 @@ class DefaultSource
       inferSchemaFlag,
       codec,
       nullValue,
-      dateFormat)(sqlContext)
+      dateFormat,
+      treatParseExceptionAsNullFlag)(sqlContext)
   }
 
   override def createRelation(
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -64,9 +64,11 @@ object TypeCast {
           case _: IntegerType => datum.toInt
           case _: LongType => datum.toLong
           case _: FloatType => Try(datum.toFloat)
-                               .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).floatValue())
+                               .getOrElse(NumberFormat
+                                          .getInstance(Locale.getDefault).parse(datum).floatValue())
           case _: DoubleType => Try(datum.toDouble)
-                                .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).doubleValue())
+                                .getOrElse(NumberFormat.getInstance(Locale.getDefault)
+                                           .parse(datum).doubleValue())
           case _: BooleanType => datum.toBoolean
           case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
           case _: TimestampType if dateFormatter != null =>
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -21,6 +21,7 @@ import java.text.{SimpleDateFormat, NumberFormat}
 import java.util.Locale
 
 import org.apache.spark.sql.types._
+import org.json4s.ParserUtil.ParseException
 
 import scala.util.Try
 
@@ -45,7 +46,8 @@ object TypeCast {
       nullable: Boolean = true,
       treatEmptyValuesAsNulls: Boolean = false,
       nullValue: String = "",
-      dateFormatter: SimpleDateFormat = null): Any = {
+      dateFormatter: SimpleDateFormat = null,
+      parseExceptionAsNull : Boolean = false): Any = {
     // if nullValue is not an empty string, don't require treatEmptyValuesAsNulls
     // to be set to true
     val nullValueIsNotEmpty = nullValue != ""
@@ -55,25 +57,32 @@ object TypeCast {
       ){
       null
     } else {
-      castType match {
-        case _: ByteType => datum.toByte
-        case _: ShortType => datum.toShort
-        case _: IntegerType => datum.toInt
-        case _: LongType => datum.toLong
-        case _: FloatType => Try(datum.toFloat)
-          .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).floatValue())
-        case _: DoubleType => Try(datum.toDouble)
-          .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).doubleValue())
-        case _: BooleanType => datum.toBoolean
-        case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
-        case _: TimestampType if dateFormatter != null =>
-          new Timestamp(dateFormatter.parse(datum).getTime)
-        case _: TimestampType => Timestamp.valueOf(datum)
-        case _: DateType if dateFormatter != null =>
-          new Date(dateFormatter.parse(datum).getTime)
-        case _: DateType => Date.valueOf(datum)
-        case _: StringType => datum
-        case _ => throw new RuntimeException(s"Unsupported type: ${castType.typeName}")
+      try {
+        castType match {
+          case _: ByteType => datum.toByte
+          case _: ShortType => datum.toShort
+          case _: IntegerType => datum.toInt
+          case _: LongType => datum.toLong
+          case _: FloatType => Try(datum.toFloat)
+                               .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).floatValue())
+          case _: DoubleType => Try(datum.toDouble)
+                                .getOrElse(NumberFormat.getInstance(Locale.getDefault).parse(datum).doubleValue())
+          case _: BooleanType => datum.toBoolean
+          case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
+          case _: TimestampType if dateFormatter != null =>
+            new Timestamp(dateFormatter.parse(datum).getTime)
+          case _: TimestampType => Timestamp.valueOf(datum)
+          case _: DateType if dateFormatter != null =>
+            new Date(dateFormatter.parse(datum).getTime)
+          case _: DateType => Date.valueOf(datum)
+          case _: StringType => datum
+          case _ => throw new UnsupportedTypeException(s"Unsupported type: ${castType.typeName}")
+        }
+      }
+      catch {
+        case e : UnsupportedTypeException =>
+          throw e
+       case e => if (parseExceptionAsNull && nullable) null else throw e
       }
     }
   }
@@ -106,3 +115,6 @@ object TypeCast {
     }
   }
 }
+
+class UnsupportedTypeException(message: String = null, cause: Throwable = null)
+  extends RuntimeException(message, cause)
--- a/src/test/scala/com/databricks/spark/csv/CsvSuite.scala
+++ b/src/test/scala/com/databricks/spark/csv/CsvSuite.scala
@@ -32,6 +32,7 @@ import org.scalatest.Matchers._
 abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
   val carsFile = "src/test/resources/cars.csv"
   val carsMalformedFile = "src/test/resources/cars-malformed.csv"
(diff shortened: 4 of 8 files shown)


HyukjinKwon:
@falaki Actually, shouldn't such behaviour be included in `PERMISSIVE` mode?

HyukjinKwon:
@rachelwarren Would you maybe share some codes and outputs after this change?

AFAIK JSON data source at Spark produces `null` in this case. For me I think maybe it does not necessarily have to work exactly as the same for each mode in data sources but I think it would be great if they are consistent.

I am trying to add parse modes in JSON at Spark, https://github.com/apache/spark/pull/11756 here.

rachelwarren:
I'm not exactly sure what you mean. The goal is that running something like the code: 
 new CsvParser().
        withUseHeader(header).
        withDelimiter(delim).
        withQuoteChar(quoteString).
        withSchema(schema).
        csvFile(sqlContext, path)

will not fail in the general case, even if the schema specifies for example a numeric column, and there is a string value in that column. 
I.e. 
If we have 
year, cars 
1995, "Toyota"
2000, "Honda"
new,  "Ford"
then this will be read in as 
1995, "Toyota"
2000, "Honda"
null,  "Ford"
if the year column is an integer.

HyukjinKwon:
Ah, I mean I think this behaviour should be included in `PERMISSIVE` mode not an option because JSON data source will produce the results in this way with this `PERMISSIVE` mode.

For example, the data below:

```json
{"a": {"b": 1}}
{"a": []}
```

with the schema below:

```scala
val schema =
  StructType(
    StructField("a", StructType(
      StructField("b", StringType) :: Nil
    )) :: Nil)
```

This produces the results below:

```bash
+----+
|   a|
+----+
| [1]|
|null|
+----+
```

rachelwarren:
i see. I can look into making that fix.

HyukjinKwon:
Maybe we should wait for feedback. That was just my opinion.

HyukjinKwon:
I realised this is actually related with https://github.com/databricks/spark-csv/issues/286.

rachelwarren:
It does. I actually did try adding it to PERMISSIVE MODE, so you can see what that looks like on this branch: https://github.com/rachelwarren/spark-csv/tree/permissive

dmsuehir:
Hello,

What's the status on this pull request?  I'm looking for this feature to get a null value when it can't be parsed.  Will it be getting in as a new option or as a modification to PERMISSIVE?  Can we expect to see the change get in master anytime soon?

Thanks,
Dian

rachelwarren:
Its still hanging out. I have implemented the permissive strategy as well.
https://github.com/rachelwarren/spark-csv/tree/permissive

joyeshmishra:
Could this be reviewed/merged? (or the permissive PR)

falaki:
I somewhat disagree with this pull request. For example, if there is a string in a field, user can either
* Use schema inference to correctly mark that field as `StringType`
* Change/fix the schema passed to data source

Would you please explain why neither of those two options work in your use case?

rachelwarren:
This addresses a dirty data issue. The columns are not string columns, they are integer or data columns, they just may have a few malformed values, when working in the real data work i run into this instance all the time. Part of the trouble is that different systems represent null values in a different way. For example a null value in some systems that we use is represented as "N/A" which is a parsed in as a string, when it should really be interpreted as a missing value. With user entered data there may be errors especially in date formats. The dataset may have many columns, I don't want to drop the entire row just because the value 5.0 is in one integer column.
The alternative to this pull request is to read in every row as strings so that it doesn't fail on dirty data and then convert them safely to the real type. This solution, in my opinion defeats the purpose of using an off the shelf CSV parser, at all (and is why we have stopped using this parser). 

@joyeshmishra @dmsuehir  Does this sound relevant to your use cases?

dmsuehir:
Yes, that is exactly why we want this feature as well.  We don't want to stop parsing when we come across a value that can't be parsed to the specified data type, and we don't want to change the schema to accommodate bad values.  It's kind of similar to the DROPMALFORMED mode, except that we don't want the entire row dropped, we just want null/missing values where the bad values were.

falaki:
I see. Thanks for the feedback. Also do you guys plan to use Spark 2.0? This data source has been inlined in Spark 2.0. So the package would only work with Spark 1.X.

rachelwarren:
We are not going to update for a while so this would be fine, but I can
write a 2.0 version in the next few weeks as well.

HyukjinKwon:
FYI, this still happens in the csv one at Spark 2.0 as well. It might be great if this issue is ported to JIRA anyway.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -21,6 +21,7 @@ import java.text.{SimpleDateFormat, NumberFormat}
 import java.util.Locale
 
 import org.apache.spark.sql.types._
+import org.json4s.ParserUtil.ParseException
```
falaki: This import is not used and not needed.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -55,25 +57,34 @@ object TypeCast {
...
-        case _: StringType => datum
-        case _ => throw new RuntimeException(s"Unsupported type: ${castType.typeName}")
+      try {
+        castType match {
+          case _: ByteType => datum.toByte
+          case _: ShortType => datum.toShort
+          case _: IntegerType => datum.toInt
+          case _: LongType => datum.toLong
+          case _: FloatType => Try(datum.toFloat)
+                               .getOrElse(NumberFormat
```
falaki: Fix indent here and bellow in this block.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -55,25 +57,34 @@ object TypeCast {
...
+          case _: TimestampType => Timestamp.valueOf(datum)
+          case _: DateType if dateFormatter != null =>
+            new Date(dateFormatter.parse(datum).getTime)
+          case _: DateType => Date.valueOf(datum)
+          case _: StringType => datum
+          case _ => throw new UnsupportedTypeException(s"Unsupported type: ${castType.typeName}")
+        }
+      }
+      catch {
+        case e : UnsupportedTypeException =>
```
falaki: Fix spacing: `case e: UnsupportedTypeException => `

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -55,25 +57,34 @@ object TypeCast {
...
+          throw e
+       case e => if (parseExceptionAsNull && nullable) null else throw e
```
falaki: `case e: Throwable => ...`

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -115,4 +117,48 @@ class TypeCastSuite extends FunSuite {
...
     assert(TypeCast.castTo("", StringType, true, true, "") == null)
   }
+
+  test("Parse exception is caught correctly"){
+
+    def testParseException( castType : DataType, badValues : Seq[String]): Unit = {
+      badValues.foreach(testValue => {
+        assert(TypeCast.castTo(testValue, castType, true, false, "", null, true) == null)
+        // if not nullable it isn't null
+        assert(Try(TypeCast.castTo(testValue, castType, false, false, "", null, true)).isFailure)
```
falaki: Instead of using `Try` catch the exception and then assert that it includes correct message.

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -115,4 +117,48 @@ class TypeCastSuite extends FunSuite {
     assert(TypeCast.castTo("", StringType, true, false, "") == "")
     assert(TypeCast.castTo("", StringType, true, true, "") == null)
   }
+
+  test("Parse exception is caught correctly"){
```
falaki: Add space `test("Parse exception is caught correctly") {`

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -115,4 +117,48 @@ class TypeCastSuite extends FunSuite {
...
+    val timestamp = "2015-01-01 00:00:00"
+    assert(TypeCast.castTo(timestamp, TimestampType, true, false, "", null, true)
+      == Timestamp.valueOf(timestamp))
+    testParseException(TimestampType, Seq("5", "string"))
+
+    assert(TypeCast.castTo("2015-01-01", DateType, true, false, "", null, true)
+      == Date.valueOf("2015-01-01"))
+    testParseException(DateType, Seq("5", "string", timestamp))
+  }
+
```
falaki: Extra lines

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -125,6 +125,16 @@ class DefaultSource
       throw new Exception("Treat empty values as null flag can be true or false")
     }
 
+    val treatParseExceptionAsNull = parameters.getOrElse(
+      "treatParseExceptionAsNull", "false")
```
falaki: I suggest changing name of this option to something like `insertNullOnErrors`. Primarily because this data source is used by R and Python users and exceptions are not familiar concepts for them.

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -125,6 +125,16 @@ class DefaultSource
       throw new Exception("Treat empty values as null flag can be true or false")
     }
 
+    val treatParseExceptionAsNull = parameters.getOrElse(
```
falaki: Please update README file to include this new option.

falaki:
I did a first pass on it. @HyukjinKwon Would you also take a look?

HyukjinKwon:
@falaki Sure! thank you for cc me.

HyukjinKwon:
@falaki, Actually, I noticed this before and have been thinking overtime (but I did not to open an issue because I thought it is a too narrow case).

Since Spark 2.0 JSON datasource also has the parse modes just like CSV, I thought it might be great if it is consistent.

Currently, it is included in JSON datasource, for example,

```scala
val rdd = spark.sparkContext.makeRDD(Seq("{\"a\" : 1}", "{\"a\" : \"a\"}"))
val schema = StructType(StructField("a", IntegerType, nullable = true) :: Nil)
spark.read.option("mode", "PERMISSIVE").schema(schema).json(rdd).show()
```

produces the results below:

```
+----+
|   a|
+----+
|   1|
|null|
+----+
```

Would this make sense? I don't mind if you feel strongly adding this as an option is correct but just I wanted to let you know just in case.

falaki:
I think for data source it makes more sense not to change behavior of existing configuration. In that case, it is safer to add a new mode.

HyukjinKwon:
I hear you. Thanks!

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -197,9 +204,44 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
...
       .collect().size
 
+    val r = new CsvParser()
+                  .withSchema(strictSchema)
+                  .withUseHeader(true)
+                  .withParserLib(parserLib)
+                  .withParseMode(ParseModes.DROP_MALFORMED_MODE)
+                  .csvFile(sqlContext, ageFile)
+                  .select("Name")
+                  .collect()
```
HyukjinKwon: We might have to fix the indentation here,

```scala
val r = new CsvParser()
  .withSchema(strictSchema)
  .withUseHeader(true)
  .withParserLib(parserLib)
  .withParseMode(ParseModes.DROP_MALFORMED_MODE)
  .csvFile(sqlContext, ageFile)
  .select("Name")
  .collect()
```

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -267,6 +309,7 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
       .withDelimiter(',')
       .withQuoteChar(null)
       .withUseHeader(true)
+      .withNullValue("")
```
HyukjinKwon: I guess this one is not related with this PR. The default value will be already `""`.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -118,7 +120,8 @@ case class CsvRelation protected[spark] (
           while (index < schemaFields.length) {
             val field = schemaFields(index)
             rowArray(index) = TypeCast.castTo(tokens(index), field.dataType, field.nullable,
-              treatEmptyValuesAsNulls, nullValue, simpleDateFormatter)
+              treatEmptyValuesAsNulls, nullValue, simpleDateFormatter,
+              treatParseExceptionAsNull)
```
HyukjinKwon: It might be great if the indentation is consistent, for example, with https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/CsvRelation.scala#L195-L202.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -55,25 +57,34 @@ object TypeCast {
...
+          case _: BooleanType => datum.toBoolean
+          case _: DecimalType => new BigDecimal(datum.replaceAll(",", ""))
+          case _: TimestampType if dateFormatter != null =>
+            new Timestamp(dateFormatter.parse(datum).getTime)
+          case _: TimestampType => Timestamp.valueOf(datum)
+          case _: DateType if dateFormatter != null =>
+            new Date(dateFormatter.parse(datum).getTime)
+          case _: DateType => Date.valueOf(datum)
+          case _: StringType => datum
+          case _ => throw new UnsupportedTypeException(s"Unsupported type: ${castType.typeName}")
```
HyukjinKwon: Could we maybe just do this, for example, as below without `try/catch`:

```scala
case _ if (parseExceptionAsNull && nullable) => null
case _ => throw new UnsupportedTypeException(s"Unsupported type: ${castType.typeName}")
```

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
HyukjinKwon: (maybe `val parser` instead of `val r` just to be consistent)

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -115,4 +117,48 @@ class TypeCastSuite extends FunSuite {
...
+  test("Parse exception is caught correctly"){
+
+    def testParseException( castType : DataType, badValues : Seq[String]): Unit = {
+      badValues.foreach(testValue => {
+        assert(TypeCast.castTo(testValue, castType, true, false, "", null, true) == null)
+        // if not nullable it isn't null
+        assert(Try(TypeCast.castTo(testValue, castType, false, false, "", null, true)).isFailure)
+      }
+      )
+    }
```
HyukjinKwon: Indentation and remove extra function here, (see, [scala-style-guide#anonymous-methods](https://github.com/databricks/scala-style-guide#anonymous-methods)) maybe as below:

```scala
def testParseException(castType : DataType, badValues : Seq[String]): Unit = {
  badValues.foreach { testValue =>
    assert(TypeCast.castTo(testValue, castType, true, false, "", null, true) == null)
    // if not nullable it isn't null
    assert(Try(TypeCast.castTo(testValue, castType, false, false, "", null, true)).isFailure)
  }
}
```

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
HyukjinKwon: Oh, navermind. I got it.

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
HyukjinKwon: Hm.. I took another look. It seems this `val r` is not related with this PR and not used?

HyukjinKwon:
Filed here, https://issues.apache.org/jira/browse/SPARK-16512

rachelwarren:
Thanks for all the feedback. I can address this next week, and hopefully get it in. Then I'll work on the 2.0 version after that.

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
rachelwarren: This could be any manor of type casting exceptions though? Its anything thrown by the type conversion.

rachelwarren:
Sorting out all the merge conflicts associated with this PR was a bit gnarly. I have made a new request and am going to close this one.

Issue #299, opened 2016-03-21
Missing documentation for treatEmptyValuesAsNulls option.


HyukjinKwon:
I can't find a description for a option `treatEmptyValuesAsNulls` in `README.md`.

Is this intentionally omitted (maybe is this deprecated)?

falaki:
I don't think it is intentional. The option is used in `TypeCast`

PR #305, opened 2016-03-22
Missing documentation for treatEmptyValuesAsNulls option.

HyukjinKwon:
https://github.com/databricks/spark-csv/issues/299

This PR adds a missing option, `treatEmptyValuesAsNulls` in documentation.

Also, this corrects the Python example for compressed output, which has an indentation different with the example right above.


HyukjinKwon:
Closing this as it is now maintenance mode and it seems it is not a critical bug.

Issue #304, opened 2016-03-22
Not implemented but exposed options (ignoreTrailingWhiteSpace and ignoreLeadingWhiteSpace)


HyukjinKwon:
It looks `ignoreTrailingWhiteSpace` and `ignoreLeadingWhiteSpace` options are exposed but not implemented.

HyukjinKwon:
I think this is related wth https://github.com/databricks/spark-csv/issues/158 if both options mean trimming heading and trailing spaces.

HyukjinKwon:
@falaki Do you intent to implement them or remove them? I will submit a PR if you allow me.

falaki:
Yes, I noticed this problem when I ported the package to Spark 2.0. I think since they are exposed in docs, it is better to implement them.

HyukjinKwon:
Let me try to create a PR first.
(Actually, I just realised they are not exposed in `README.md` but just as parameters in some functions)

dereXeus:
Hi guys, any idea how can I implement this ?
I see that new Parser is being made with settings : 
`new CsvParser(settings)`
and in settings :
`settings.setIgnoreLeadingWhitespaces(ignoreLeadingSpace)`
and default value is :
`ignoreLeadingSpace: Boolean = true`

So when using Univocity , this should be enabled by default ?

PR #306, opened 2016-03-23
Add the support for ignoreTrailingWhiteSpace and ignoreLeadingWhiteSpace

HyukjinKwon:
https://github.com/databricks/spark-csv/issues/304

This PR adds the support for `ignoreTrailingWhiteSpace` and `ignoreLeadingWhiteSpace` options which were exposed but not implemented.

For Apache Common CSV Parser (`common`),  it supports the functionalities only when both `ignoreTrailingWhiteSpace` and `ignoreLeadingWhiteSpace` are set to `true` by using `withIgnoreSurroundingSpaces()` at `CSVFormat`.

For Univocity parser (`univocity`),  it supports this functionalities for `ignoreTrailingWhiteSpace`, and/or `ignoreLeadingWhiteSpace` by using `setIgnoreLeadingWhitespaces()` and/or `setIgnoreTrailingWhitespaces()` at `CsvParserSettings`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ The package also supports saving simple (non-nested) DataFrame. When writing fil
 * `nullValue`: specificy a string that indicates a null value, nulls in the DataFrame will be written as this string.
 * `codec`: compression codec to use when saving to file. Should be the fully qualified name of a class implementing `org.apache.hadoop.io.compress.CompressionCodec` or one of case-insensitive shorten names (`bzip2`, `gzip`, `lz4`, and `snappy`). Defaults to no compression when a codec is not specified.
 * `quoteMode`: when to quote fields (`ALL`, `MINIMAL` (default), `NON_NUMERIC`, `NONE`), see [Quote Modes](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/QuoteMode.html)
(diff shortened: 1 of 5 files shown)


dbspace:
I try to use this, but I am facing a strange issue. I am able to create a dataframe from the csv, but if I try to use the dataframe that is created and convert 1 of column from string to date using withColumn and then try to write it out to parquet I get a number format exception. However if I write the dataframe out as parquet first and read the parquet back and use the withColumn to create a new dataframe and write that out everything works fine.  I can create a test case if you need more information

HyukjinKwon:
I can rebase this if it seems worth adding it.

mrchristine:
Any reason users would want to keep whitespace on one side? I'd recommend simplifying this to add "ignoreHeaderWhitespace" to ignore trailing and leading whitespace. I don't see a use case where users would want to keep one or the other, and it makes referring to columns more complex. Users might run into more issues if they only add 1 option, but wanted both.

HyukjinKwon:
Closing this as it is now maintenance mode and it seems it is not a critical bug.

rvisio:
Any plans to add this?  I noticed the two ignore whitespace options on the javadoc but was unable to use them

dereXeus:
I have forked this repo. Any chances you can raise a pull request to it ?

PR #307, opened 2016-03-24
Allow for maxCharsPerCol to be set

addisonj:
Currently, the maxCharactersPerColumn value is hardcoded to
100,000. I (sadly) have CSV fields with more than 100k characters so
need to be able to configure this.

This allows the value to be passed the same way as other paramters while
keeping a backward compatible default of 100k


falaki:
Thanks for submitting this. Please add the option to `package.scala` so it can be set from SQL options as well.

addisonj:
Done! Thanks for the quick response!

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -287,7 +288,8 @@ case class CsvRelation protected[spark] (
 
         new BulkCsvReader(iter, split,
           headers = header, fieldSep = delimiter,
-          quote = quoteChar, escape = escapeVal, commentMarker = commentChar)
+          quote = quoteChar, escape = escapeVal,
+          commentMarker = commentChar, maxCharsPerCol = maxCharsPerCol)
```
HyukjinKwon: Is this `univocity` parser only option?

src/test/scala/com/databricks/spark/csv/CsvSuite.scala
```diff
@@ -852,6 +853,38 @@ abstract class AbstractCsvSuite extends FunSuite with BeforeAndAfterAll {
     assert(results.toSeq.map(_.toSeq) === expected)
   }
 
+  test("DSL allows for setting maxColsPerChar and expect error") {
+    val parser = new CsvParser()
+      .withDelimiter(',')
+      .withUseHeader(true)
+      .withParseMode(ParseModes.FAIL_FAST_MODE)
+      .withMaxCharsPerCol(5000)
```
HyukjinKwon: If this option is only for `univocity`, it would be better if we test this only when `parseLib` is `univocity`, which can be like this:
```scala
test("DSL allows for setting maxColsPerChar and expect error") {
  assume(parserLib == "UNIVOCITY", "maxCharsPerCol option is only " +
    "for univocity parser.")
  val parser = new CsvParser()
...
```

HyukjinKwon:
I think we might also need a description of `maxCharsPerCol` in `README.md`.

BTW, how does this work for Apache Common CSV parser?

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -142,6 +142,14 @@ class DefaultSource
 
     val codec = parameters.getOrElse("codec", null)
 
+    val maxCharsPerColStr = parameters.getOrElse("maxCharsPerCol", "100000")
+    val maxCharsPerCol = try {
+      maxCharsPerColStr.toInt
+    } catch {
+      case e: Exception => throw new Exception("maxCharsPerCol must be a valid integer")
+    }
+
```
HyukjinKwon: Maybe single white space if you are going to add some more commits?

HyukjinKwon:
I think we might also need a description of `maxCharsPerCol` in `README.md`.

BTW, how does this work for Apache Common CSV parser?

src/main/scala/com/databricks/spark/csv/DefaultSource.scala
```diff
@@ -142,6 +142,14 @@ class DefaultSource
...
+
```
HyukjinKwon: Maybe single white space here if you are going to add some more commits? (See [scala-style-guide#blank-lines-vertical-whitespace](https://github.com/databricks/scala-style-guide#blank-lines-vertical-whitespace))

HyukjinKwon:
I think it might also need a description of `maxCharsPerCol` in `README.md`.

BTW, how does this work for Apache Common CSV parser?

HyukjinKwon:
Also I also noticed that this feature is added in Spark 2.0, [here](https://github.com/apache/spark/blob/e474088144cdd2632cf2fef6b2cf10b3cd191c23/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVOptions.scala#L96). I think it might be good if it has the same option name and variable name.

srowen:
This sounds like https://issues.apache.org/jira/browse/SPARK-14103 ; to be clear, this PR isn't necessary for the Spark copy of this code right?

falaki:
Yes, this is already resolved in Spark 2.0.

Issue #308, opened 2016-03-24
dateFormat should be used when writing


barrybecker4:
This is related to #279

I don't think that the dateFormat specified in the options is used when writing.
Here is what I am doing and it does not write in the specified format:
repartitionDf.write.format("com.databricks.spark.csv")
        .option("delimiter", "\t")
        .option("header", "false")
        .option("nullValue", "?")    // this worked
        .option("dateFormat", "MM/dd/yyyy HH:mm:ss")
        .save(tempFileName)
The file has dates like "2012-01-03 09:12:00.0" but it should be 03/01/2012 09:12:00"

Also, I think it might be risky to use SimpleDateFormat rather than joda's DateTimeFormat, but that could be separate (and I am not sure).

HyukjinKwon:
Indeed. I think it might need to support to write as well. Could I maybe ask why it might be risky?

HyukjinKwon:
Oh, I just checked https://github.com/databricks/spark-csv/issues/279#issuecomment-201036872. It would be okay. AKAIK, That `SimpleDateFormat` object would be created in driver-side once but will be serialised and sent to each task in executer-side, meaning each object in each thread. I think I should double-check though.

HyukjinKwon:
Maybe you can try this code just to double-check. Although it is a bit messy, I think this shows correctly that `simpleDateFormat` object does not affect each task.

```scala
val path = "any-text-file.txt"
val simpleDateFormat = new SimpleDateFormat()
simpleDateFormat.setTimeZone(TimeZone.getTimeZone("Asia/Seoul"))
println(s"Driver-side : ${simpleDateFormat.getTimeZone.getID}")
sparkContext.textFile(path).repartition(3).foreach { _ =>
  println(s"Excuter-side : ${simpleDateFormat.getTimeZone.getID}")
  simpleDateFormat.setTimeZone(TimeZone.getTimeZone("Europe/Copenhagen"))
} 
println(s"Driver-side : ${simpleDateFormat.getTimeZone.getID}")
```

`Excuter-side : Asia/Seoul` will be printed for each partition (can be checked by varying `repartition(#)`, meaning modification in a thread does not affect another.
`Driver-side : Asia/Seoul` will be printed at the end, meaning modification in threads does not affect driver-side.

HyukjinKwon:
@falaki I think I could submit a PR for this if you think it needs.

barrybecker4:
[...]

PR #310, opened 2016-03-31
Fix for #308 write using specified dateFormat

barrybecker4:
Added support for writing to file with dateFormat that is specified in the options.  Added some basic tests for it.

I'm continuing to use SimpleDateFormat since that is what is used for reading. I would prefer joda's DateTimeFormat, but that can be a separate issue.

It's too bad CSVFormat does not allow specification of a dateFormat as it does for the null value. That would make this much easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ When reading files the API accepts several options:
 * `inferSchema`: automatically infers column types. It requires one extra pass over the data and is false by default
 * `comment`: skip lines beginning with this character. Default is `"#"`. Disable comments by setting this to `null`.
 * `nullValue`: specifies a string that indicates a null value, any fields matching this string will be set as nulls in the DataFrame
-* `dateFormat`: specifies a string that indicates a date format to use when reading or writing dates or timestamps. Custom date formats follow the formats at [`java.text.SimpleDateFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html). This applies to both `DateType` and `TimestampType`. By default, it is `null` which means trying to parse times and date by `java.sql.Timestamp.valueOf()` and `java.sql.Date.valueOf()`. If no dateFormat is specified, then "yyyy-MM-dd HH:mm:ss.S" is used for writing.
+* `dateFormat`: specifies a string that indicates the date format to use when reading dates or timestamps. Custom date formats follow the formats at [`java.text.SimpleDateFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html). This applies to both `DateType` and `TimestampType`. By default, it is `null` which means trying to parse times and date by `java.sql.Timestamp.valueOf()` and `java.sql.Date.valueOf()`.
 
 The package also supports saving simple (non-nested) DataFrame. When writing files the API accepts several options:
 * `path`: location of files.
@@ -65,6 +65,7 @@ The package also supports saving simple (non-nested) DataFrame. When writing fil
 * `quote`: by default the quote character is `"`, but can be set to any character. This is written according to `quoteMode`.
 * `escape`: by default the escape character is `\`, but can be set to any character. Escaped quote characters are written.
 * `nullValue`: specifies a string that indicates a null value, nulls in the DataFrame will be written as this string.
+* `dateFormat`: specifies a string that indicates the date format to use writing dates or timestamps. Custom date formats follow the formats at [`java.text.SimpleDateFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html). This applies to both `DateType` and `TimestampType`. If no dateFormat is specified, then "yyyy-MM-dd HH:mm:ss.S".
 * `codec`: compression codec to use when saving to file. Should be the fully qualified name of a class implementing `org.apache.hadoop.io.compress.CompressionCodec` or one of case-insensitive shorten names (`bzip2`, `gzip`, `lz4`, and `snappy`). Defaults to no compression when a codec is not specified.
 * `quoteMode`: when to quote fields (`ALL`, `MINIMAL` (default), `NON_NUMERIC`, `NONE`), see [Quote Modes](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/QuoteMode.html)
 
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ When reading files the API accepts several options:
 * `charset`: defaults to 'UTF-8' but can be set to other valid charset names
 * `inferSchema`: automatically infers column types. It requires one extra pass over the data and is false by default
 * `comment`: skip lines beginning with this character. Default is `"#"`. Disable comments by setting this to `null`.
-* `nullValue`: specificy a string that indicates a null value, any fields matching this string will be set as nulls in the DataFrame
-* `dateFormat`: specificy a string that indicates a date format. Custom date formats follow the formats at [`java.text.SimpleDateFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html). This applies to both `DateType` and `TimestampType`. By default, it is `null` which means trying to parse times and date by `java.sql.Timestamp.valueOf()` and `java.sql.Date.valueOf()`.
+* `nullValue`: specifies a string that indicates a null value, any fields matching this string will be set as nulls in the DataFrame
+* `dateFormat`: specifies a string that indicates a date format to use when reading or writing dates or timestamps. Custom date formats follow the formats at [`java.text.SimpleDateFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html). This applies to both `DateType` and `TimestampType`. By default, it is `null` which means trying to parse times and date by `java.sql.Timestamp.valueOf()` and `java.sql.Date.valueOf()`. If no dateFormat is specified, then "yyyy-MM-dd HH:mm:ss.S" is used for writing.
 
 The package also supports saving simple (non-nested) DataFrame. When writing files the API accepts several options:
 * `path`: location of files.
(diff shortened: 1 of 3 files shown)


falaki:
Thanks for submitting this. I am going to review it. Meanwhile would you please take a look at the test failures?

barrybecker4:
When I compile and run locally it works, but I am using spark 1.6. Not sure what the build server is using. It complains that write is not a member on DataFrame, but it is in 1.6,

HyukjinKwon:
@barrybecker4 For me, I usually run this tests below:

```bash
sbt -Dspark.testVersion=1.3.0 test &&\
sbt -Dspark.testVersion=1.4.1 test &&\
sbt -Dspark.testVersion=1.6.0 test &&\
sbt scalastyle && \
sbt "test:scalastyle"
```

You can check why it fails in [travis-ci](https://travis-ci.org/databricks/spark-csv/jobs/119900277).

It looks the reasons are below:

```bash
[error] /home/travis/build/databricks/spark-csv/src/main/scala/com/databricks/spark/csv/package.scala:170:0: Whitespace at end of line
[error] /home/travis/build/databricks/spark-csv/src/main/scala/com/databricks/spark/csv/package.scala:103: File line length exceeds 100 characters
```

barrybecker4:
I fixed the two style errors, and all those tests pass except for the last sbt "test:scalastyle" one.
It reports:

<file name="C:\work\spark-csv\src\test\scala\com\databricks\spark\csv\CsvSuite.scala"> <error severity="error" message="expected start of definition, but was Token(VAL,val,2967,val)"/>

but does not say on what line, so I don't know how to fix it. Its also not in a file that I modified.

HyukjinKwon:
@barrybecker4 It looks your tests are using `DataFrame.write()` function which (AFAIK) does not exist in Spark 1.3.x. Maybe you should just follow the function usages in `CSVSuite.scala`.

barrybecker4:
I will look at using the alternative method to write. Is it really necessary to support all the way back to 1.3.x.?

barrybecker4:
I switched to using dates.saveAsCsvFile instead of dates.write so it will be compatible with spark 1.3.0.

I am seeing this stylecheck error in CsvSuite, but I did not change anything in that file.
spark\csv\CsvSuite.scala: expected start of definition, but was Token(VAL,val,2967,val)

HyukjinKwon:
@falaki I think it is ready to be reviewed.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -161,10 +169,16 @@ package object csv {
 
           override def next: String = {
             if (iter.nonEmpty) {
-              val row = csvFormat.format(iter.next().toSeq.map(_.asInstanceOf[AnyRef]): _*)
+              val values: Seq[AnyRef] = iter.next().toSeq.map(
```
falaki: Do have a sense of the overhead of this additional type checking for each record? If the overhead is substantial, maybe we can check the schema for existence of `TimestampType` and/or `DateType` and if we find one, then we can perform the proper formatting. Even better would be using the index of the field instead of type check when we know there are Date or Time types.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -97,6 +100,11 @@ package object csv {
                       compressionCodec: Class[_ <: CompressionCodec] = null): Unit = {
       // TODO(hossein): For nested types, we may want to perform special work
       val delimiter = parameters.getOrElse("delimiter", ",")
+      // Before this change the csvFormatter wrote dates like this:
+      // "2014-11-15 06:31:10.0", so have that as the default.
+      val dateFormat = parameters.getOrElse("dateFormat", "yyyy-MM-dd HH:mm:ss.S")
```
falaki: Would be good to add this to the README file.

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
```diff
@@ -0,0 +1,129 @@
...
+  test("Save with default date format") {
+    mkTempDir()
+    val dates = readDatesFromFile()
+    val copyFilePath = tempEmptyDir + "dates-copy.csv"
+    val retDataFile = tempEmptyDir + "dates-result.csv"
+
+    // Write dataframe this way prior in spark 1.3 and before
+    dates.saveAsCsvFile(copyFilePath, Map("header" -> "false"))
+
+    // This method of writing can be used after spark 1.3
```
falaki: Check out existing tests. You can use `assume` to ignore a specific test for older versions of Spark.

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
```diff
@@ -0,0 +1,129 @@
...
+    dates.show()
+
+    val copyFilePath = tempEmptyDir + "dates-copy.csv"
+    val retDataFile = tempEmptyDir + "dates-result.csv"
+
+    // Write dataframe this way prior in spark 1.3 and before
+    dates.saveAsCsvFile(copyFilePath,
+      Map("header" -> "false", "dateFormat" -> "MM/dd/yyyy HH:mm:ss"))
+
+    // This method of writing can be used after spark 1.3
```
falaki: Ditto

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
```diff
@@ -0,0 +1,129 @@
...
+import org.scalatest.{BeforeAndAfterAll, FunSuite}
+
+import scala.io.Source
+
+class CsvWriteSuite extends FunSuite with BeforeAndAfterAll {
+
+  val datesFile = "src/test/resources/dates.csv"
+  val tempEmptyDir = "target/test/empty/"
+
+  def parserLib: String = "COMMONS"
```
falaki: There reason we had this function in `CsvSuite` was to write tests once and run for two different parser libs. Would be good to make this Suite an abstract class and extend it twice for each parser lib.

falaki:
@barrybecker4 thanks for fixing tests. I left a few comments. My main concern is the performance bottleneck inside the iterator. If you think there is risk of performance regression because of the additional type checking, would you change it to use schema information?

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
barrybecker4: Are the parser libs used when writing? These tests are used to test writing, not reading of the file. It does not matter which reader is used (as long as it works). Testing the reader with different parser libs is the job of CsvSuite not CsvWriteSuite. So is this subclassing necessary?

barrybecker4:
Are there any performance tests that indicate that it is significantly slower? I have not done any such tests yet. How much degradation is acceptable 0%, 5%, 10%?

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
falaki: Only "commons" is used for writing, but since you are reading data as well, it is safer to test with both parsers.

falaki:
We don't have any perf tests for spar-csv. Would you mind trying it manually and investigate?

HyukjinKwon:
@barrybecker4 I think we don't have to check if every value is `Timestamp` or `Date` type for every time since the data types of rows would be consistent as just @falaki said.

Isn't it obvious that checking only once would be faster than checking every value? I think we don't even need to test.

HyukjinKwon:
I cannot assert but this test will decrease the elapsed time by almost 20%

```scala
def time[R](block: => R): R = {
  val t0 = System.nanoTime()
  val result = block    // call-by-name
  val t1 = System.nanoTime()
  println("Elapsed time: " + (t1 - t0) + "ns")
  result
}

def test(v: Any) = {
  v match {
    case v: String => v
    case v: Int => v
    case _ => v
  }
}

val data = 0 to 10000000
time(data.foreach(v => test(v)))
time(data.foreach(v => v))
```

src/main/scala/com/databricks/spark/csv/package.scala
barrybecker4: I changed to using an index they way you suggested. I ran some ad hoc tests using a 500k row dataset and saw the time go from about 5.5s to 4.9s (about 11% faster). There were 10 columns, and only 1 was a date column.

src/main/scala/com/databricks/spark/csv/package.scala
barrybecker4: Added some extra info for the dateFormat option in the README.

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
barrybecker4: I just removed the commented code. Since it now works with 1.3, I don't think its necessary to use "assume".

src/test/scala/com/databricks/spark/csv/CsvWriteSuite.scala
barrybecker4: I made the change, but I still do not agree that the tests for writing (which do not involve the parserLib other than to read a file so that it can test the writing of it) need to be run this way. Maybe I don't understand. Are you saying that the parserLib is used for writing in addition to reading? Given the name, I thought it would only be involved with reading.

src/main/scala/com/databricks/spark/csv/package.scala
HyukjinKwon: I think he meant adding this in documentation for wring option part.

barrybecker4: I guess I don't know what the wring option part is.

README.md
```diff
@@ -55,16 +55,16 @@ When reading files the API accepts several options:
...
 
 The package also supports saving simple (non-nested) DataFrame. When writing files the API accepts several options:
 * `path`: location of files.
 * `header`: when set to true, the header (from the schema in the DataFrame) will be written at the first line.
 * `delimiter`: by default columns are delimited using `,`, but delimiter can be set to any character
 * `quote`: by default the quote character is `"`, but can be set to any character. This is written according to `quoteMode`.
 * `escape`: by default the escape character is `\`, but can be set to any character. Escaped quote characters are written.
-* `nullValue`: specificy a string that indicates a null value, nulls in the DataFrame will be written as this string.
+* `nullValue`: specifies a string that indicates a null value, nulls in the DataFrame will be written as this string.
 * `codec`: compression codec to use when saving to file. Should be the fully qualified name of a class implementing `org.apache.hadoop.io.compress.CompressionCodec` or one of case-insensitive shorten names (`bzip2`, `gzip`, `lz4`, and `snappy`). Defaults to no compression when a codec is not specified.
```
HyukjinKwon: @barrybecker4 Maybe here we could add a description for what happens if `dateFormat` is given as an option when writing.

src/main/scala/com/databricks/spark/csv/package.scala
barrybecker4: OK. I Added dateFormat property documentation to both the reading and writing sections.
Is there anything else, or can this be merged now?

barrybecker4:
Are there any more changes that I need to make before this can be merged?

barrybecker4:
It's been a couple more months, and there have been no further comments. @HyukjinKwon  or @falaki, can this pull request be merged to master?

falaki:
@barrybecker4 Looks good. I am merging this.

BTW: I am waiting for Spark 2.0 QA to conclude to make the final release of this package.

barrybecker4:
Unfortunately, I believe I just found a bug in the code that was merged. It fails if there is a null date in the date data. I will make a fix on my fork. Not sure what the process is. Do I reopen this issue or create a separate one?

falaki:
I suggest you open a new issue and submit a new PR. This is just in master, since we have not published a release yet.

Issue #317, opened 2016-04-26
Is it possible to only write one header total (rather than per file)


abridgett:
When writing out from spark to HDFS, depending upon the header setting, each file has a header.  So when calling hdfs -getmerge you get duplicated headers.  So ATM we dump without headers, then -getmerge and stick a header on manually (ugh!).   Unless we coalesce(1) which has a bad performance impact

This maybe closed as a "sorry, not possible" of course as it basically requires special casing partition 1.

HyukjinKwon:
cc @falaki Maybe it would be great if we can have your thoughts if you don't mind.

falaki:
I think it is totally possible. I prefer if we don't change the default behavior. For example, with an additional flag, e.g., `headerOnFirstFile = true`, that user can control we add this functionality. As you pointed out it is critical to be able to read it back correctly. Maybe same flag can be used when reading it.

roywei:
Hi, is there any way to manually achieve this? will mapPartitionWithIndex  work?  thanks

hangingman:
I would like to have the option `headerOnFirstFile = true`

PR #318, opened 2016-04-27
only write CSV header on the first partition (#317)

abridgett:
This seems to fix the problem I saw mentioned in #317.  I'm most certainly _not_ saying that this change should go in without someone who knows what they are doing taking a look!

In particular if anyone had been relying on the old behaviour they would get a nasty surprise.  However I think this is a more useful default.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/package.scala
+++ b/src/main/scala/com/databricks/spark/csv/package.scala
@@ -155,7 +155,7 @@ package object csv {
           .withNullString(nullValue)
 
         new Iterator[String] {
-          var firstRow: Boolean = generateHeader
+          var firstRow: Boolean = generateHeader && (index==0)
 
           override def hasNext: Boolean = iter.hasNext || firstRow
 


HyukjinKwon:
@abridgett Is this guaranteed that you can load back the written with this library with `header` option `true`? Can you try and add a test for this?

abridgett:
it probably depends where you read it from - I suspect if you read it before aggregation (i.e. getmerge) then you will lose records as I suspect that the "old" behaviour drops the first row of all files (untested).

But then the old method makes it very hard to get useful CSVs out (without setting coalesce to "1" with the performance impact that entails) or post processing to remove all but one header from the PART files. 

I'd hope that most people writing CSVs are using them in external systems (which only want one header) rather than using them in HDFS for further processing.

Re tests: I'd _love_ to, but my Scala is utterly non-existent I'm afraid and it'd probably take me weeks (not kidding!)

HyukjinKwon:
Then, I think the whole library should be changed. It does drops a header for each file during reading each. 

Spark has some data sources such as Parquet and ORC (file formats) and each part-file is a complete format and thus each part-file has `.orc` or `.parquet` extensions. In Spark internal CSV data source, each is a consistent format with `.csv` extension (this library is not). In this way, I also think there should be consistent part-files for this library as well. XML data source also writes files in this way.

Also, if we drop the header, then we also will not be able to read a directory with individual CSV files which have the same formats.

For example,

```bash
args_data/
├── ages1.csv
├── ages2.csv
└── ages3.csv
```

Also, Spark supports partitioned table which has a hierarchy like the above. Although this library does not support that (Spark internal CSV data source does), I guess we cannot support to read partitioned files.

To cut it short, I think we need to support to read CSV files written by this library. Writing inconsistent part-files will cause not supporting to read them back.

HyukjinKwon:
@abridgett How about closing this for now and discuss a proper way to fix in the issue you opened?

abridgett:
Absolutely (#317?) I wonder if that would involve making headers a tri-valued option?

HyukjinKwon:
@abridgett  I think that might be a way maybe. I think it would be nicer if we cc who has the right to decide and ask our suggestions. Let's close this and cc someone to discuss if you think it makes sense.

abridgett:
Agreed.  Would you like me to do something (as I've raised #317 already).

PR #331, opened 2016-05-14
 Spelling Correction

Mayank-Shete:
Changed  whitesspaces to white space


HyukjinKwon:
@Mayank-Shete It seems both "white space" and "whitespace" are acceptable. Is there a specific reason to not use "whitespace"?

Mayank-Shete:
@HyukjinKwon in the first instance there is a typo whitesspace and for the other part changing whitespace to white space is for code uniformity . It could be done other way around white space to whitespace

HyukjinKwon:
Ah, I see. Thanks.

Issue #356, opened 2016-07-11
support multiple dateFormats within same file


bruce-szalwinski:
I see that if my dates and timestamps are in Date.valueOf() and Timestamp.valueOf() respectively, then I don't have to specify dateFormat.  I also see that if they are not in this format, then I can use the dateFormat option to specify the format.  What I'm running into is a file that contains both date fields and timestamp fields and neither one is in the standard format.  I can specify the format for one or the other, but I get parsing errors for the date field if I specify the timestamp format and i get invalid data for the timestamp field if I specify the date format.

Is there a way to specify multiple dateFormats?

This shows the parsing error when specifying the dateFormat for the timestamp.

`import org.apache.spark.sql.types.{StructType, StructField, StringType, IntegerType, DateType, TimestampType}

val schema = StructType(Array(
     StructField("id", IntegerType, true),
     StructField("dateField", DateType, true),
     StructField("timestampField", TimestampType, true)))
     
val df = sqlContext.read.
	format("com.databricks.spark.csv").
	schema(schema).
	option("delimiter", "|").
	option("dateFormat", "yyyy-MM-dd HH:mm:ss").
	option("inferSchema", "true").
	load("./date_and_timestamp.csv")`

And given input file that looks like:

`1|01-06-2016|2016-06-01 01:01:01`

I get exception:

`scala> df.show()
16/07/11 12:52:57 ERROR Executor: Exception in task 0.0 in stage 0.0 (TID 0)
java.text.ParseException: Unparseable date: "01-06-2016"
	at java.text.DateFormat.parse(DateFormat.java:366)
	at com.databricks.spark.csv.util.TypeCast$.castTo(TypeCast.scala:73)
	at com.databricks.spark.csv.CsvRelation$$anonfun$buildScan$2.apply(CsvRelation.scala:120)
	at com.databricks.spark.csv.CsvRelation$$anonfun$buildScan$2.apply(CsvRelation.scala:106)
	at scala.collection.Iterator$$anon$13.hasNext(Iterator.scala:371)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:327)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:327)
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:308)
	at scala.collection.Iterator$class.foreach(Iterator.scala:727)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
[...]

PR #359, opened 2016-07-14
Support multiple date formats

bruce-szalwinski:
#356

This PR adds support for parsing multiple date formats with the same file.  Simply pass a list of date format strings:

```
  .withDateFormat("dd/MM/yyyy HH:mm", "yyyy/MM/dd", "yyyy-MM-dd")
```

The first one that successfully parses will be used.


falaki:
How would it look like in SQL?

bruce-szalwinski:
Good point, forgot about that case.  I see that the options clause for create table takes name/value pairs.  Would it be possible to list multiple dateFormat entries and each one adds to the list of available formatters?  It is not a case is covered by current PR.

    options(dateFormat "yyyy-mm-dd", dateFormat "yyyy/MM/dd", ...)

HyukjinKwon:
@bruce-szalwinski No, it does not support find of list or seq. (See, [SqlBase.g4](https://github.com/apache/spark/blob/34283de160808324da02964cd5dc5df80e59ae71/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4)).

For SQL, currently, only string, double (decimal), long and boolean are supported for 2.0 and only string for lower Spark versions.

Added options will eventually become `Map<String, String>`. So, only the last value will remain in the case above.

For Scala, it also only supports string, double (decimal), long and boolean are supported for 2.0 and only string for lower Spark versions.

README.md
```diff
@@ -138,6 +138,19 @@ selectedData.write
...
+You can specify one or more date formats when reading data:
+```scala
+import org.apache.spark.sql.SQLContext
+
+val sqlContext = new SQLContext(sc)
+
+val df = sqlContext.read
+    .format("com.databricks.spark.csv")
+    .option("header", "true")
+    .option("dateFormat", "dd/MM/yyyy HH:mm", "yyyy/MM/dd", "yyyy-MM-dd")
```
HyukjinKwon: Therefore, this case would not work because `option()` takes only two arguments.

bruce-szalwinski:
So, ironically, i would need a comma separate list of dateFormat strings to parse multiple date formats inside of a CSV file :).  

I should rejigger this to take:

    .option("dateFormat", "dd/MM/yyyy HH:mm, yyyy/MM/dd, yyyy-MM-dd")

perhaps?

falaki:
I suggest using multiple keys like this:
```
.option("dateFormat1", "dd/MM/yyyy HH:mm")
.option("dateFormat2", "yyyy/MM/dd, yyyy-MM-dd")
```

You need to handle the default case as well:
```
.option("dateFormat", "yyyy/MM/dd, yyyy-MM-dd")
```

bruce-szalwinski:
Thanks for the feedback.  I have incorporated it into the latest PR.  I'm running into an issue with one of the test cases where it passes in Intellij but fails from command line.  Any pointers on how to debug this?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -53,7 +53,9 @@ case class CsvRelation protected[spark] (
   extends BaseRelation with TableScan with PrunedScan with InsertableRelation {
 
   // Share date format object as it is expensive to parse date pattern.
-  private val dateFormatter = if (dateFormat != null) new SimpleDateFormat(dateFormat) else null
+  private val dateFormatter = if (dateFormat != null) {
+    dateFormat.split(",").map(df => new SimpleDateFormat(df))
+  } else null
```
HyukjinKwon: I hope the indentation is fixed as below:

```scala
if (dateFormat != null) {
  dateFormat.split(",").map(df => new SimpleDateFormat(df))
} else {
  null
}
```

and maybe it might better avoid to use name `df` because it is used for `DataFrame` as far as I know.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -46,7 +46,7 @@ package object csv {
         ignoreLeadingWhiteSpace: Boolean = false,
         ignoreTrailingWhiteSpace: Boolean = false,
         charset: String = TextFile.DEFAULT_CHARSET.name(),
-        inferSchema: Boolean = false): DataFrame = {
+        inferSchema: Boolean = false) : DataFrame = {
```
HyukjinKwon: It seems it does not have to insert a space here :)

HyukjinKwon:
Actually, there is a scala coding guilds here, https://github.com/databricks/scala-style-guide. This might be helpful.

bruce-szalwinski:
Thanks for the pointer to the style-guide.  I have updated accordingly.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -36,7 +36,7 @@ private[csv] object InferSchema {
       tokenRdd: RDD[Array[String]],
       header: Array[String],
       nullValue: String = "",
-      dateFormatter: SimpleDateFormat = null): StructType = {
+      dateFormatter: Seq[SimpleDateFormat] = null): StructType = {
```
HyukjinKwon: It would be better if you use `Seq.empty[SimpleDateFormat]` rather than `null`.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -78,7 +78,7 @@ private[csv] object InferSchema {
   private[csv] def inferField(typeSoFar: DataType,
       field: String,
       nullValue: String = "",
-      dateFormatter: SimpleDateFormat = null): DataType = {
+      dateFormatter: Seq[SimpleDateFormat] = null): DataType = {
```
HyukjinKwon: Here as well.

HyukjinKwon:
@bruce-szalwinski Do you mind if I ask how it is related with writing? as far as I remember writing also supports `dateFormat`.

bruce-szalwinski:
Yes, writing also supports `dateFormat` using the `dateFormat` option.  The user can override the default date format using the `dateFormat` option when saving a file.   All dates will be written using the single date format.

On the reader side:

    val df = sqlContext.read
        .option("dateFormat", "pattern1,pattern2,...") 
        .load("cars.csv")

On the writer side:

    val selectedData = df.select("year", "model")
    selectedData.write
        .option("dateFormat", "pattern1")
        .save("newcars.csv")

The `dateFormatters` is constructed by splitting the `dateFormat` on the comma, so the change to the dateFormat option would support both reader and writer.

bruce-szalwinski:
Anything else that I need to do here?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -53,7 +53,12 @@ case class CsvRelation protected[spark] (
   extends BaseRelation with TableScan with PrunedScan with InsertableRelation {
 
   // Share date format object as it is expensive to parse date pattern.
-  private val dateFormatter = if (dateFormat != null) new SimpleDateFormat(dateFormat) else null
+  private val dateFormatter: Seq[SimpleDateFormat] =
+    if (dateFormat != null) {
+      dateFormat.split(",").map(pattern => new SimpleDateFormat(pattern))
```
falaki: I still don't like this. What if date format contains `,`?

bruce-szalwinski: I can think of a couple of choices here.  Let me know if you have a preference.   Perhaps option b helps to preserve binary compatibility so would be favored?

a - Have another option, dateFormatDelimiter, defaults to ",".  
b - Have delimiter as first part of dateFormat.  Something like "//[delimiter]\n", ala String Calculator kata, http://osherove.com/tdd-kata-1/.  If there is no //[delimiter]\n", the delimiter defaults to ","

falaki: I suggest supporting up to three additional options: `dateFormat1`, `dateFormat2`, `dateFormat3` (including the old `dateFormat`) and do not try to tokenize the string passed by user.

PR #360, opened 2016-07-15
fix NPE that can occur for null dates

barrybecker4:
I know #310 was recently merged and closed, but I just made an additional change to fix a problem I noted when null dates are in the data. A check for null needs to be done. Let me know if I need to create a separate issue or if you can just merge this.


barrybecker4:
I created #361 for this issue since #310 was closed.

falaki:
this PR includes all the commits that are already merged in master.

barrybecker4:
Yes, I synched with the master and merged them in. Wasn't I supposed to? Not sure what I am supposed to do now. In any case, this is the change I would like on master:
https://github.com/databricks/spark-csv/pull/360/commits/c4e09b145a8b1dc07d986e2c1d415cbadd4035a8

falaki:
@barrybecker4 tests are failing.

barrybecker4:
OK. Fixed the tests. When I modified the data file, I needed to fix the read tests and not just the write tests. BTW, when I run the tests locally I see some failures when I run the CsvSuite tests. The errors are all of the form
{code}
Failed to delete: C:\work\spark-csv\target\test\empty
java.io.IOException: Failed to delete: C:\work\spark-csv\target\test\empty
{code}
It probably has something to do with running on windows.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -159,8 +159,10 @@ package object csv {
       val schema = dataFrame.schema
       val formatForIdx = schema.fieldNames.map(fname => schema(fname).dataType match {
         case TimestampType => (timestamp: Any) =>
-          dateFormatter.format(new Date(timestamp.asInstanceOf[Timestamp].getTime))
-        case DateType => (date: Any) => dateFormatter.format(date)
+          if (timestamp == null) nullValue
```
falaki: style nit: When if else block does not fit in one one, use brackets.

barrybecker4:
OK. Added the braces.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -158,9 +158,13 @@ package object csv {
       // does not have to happen in the inner loop.
       val schema = dataFrame.schema
       val formatForIdx = schema.fieldNames.map(fname => schema(fname).dataType match {
-        case TimestampType => (timestamp: Any) =>
-          dateFormatter.format(new Date(timestamp.asInstanceOf[Timestamp].getTime))
-        case DateType => (date: Any) => dateFormatter.format(date)
+        case TimestampType => (timestamp: Any) => {
+          if (timestamp == null) nullValue
```
HyukjinKwon: Is this safe? It seems the type is expected as `Timestamp` but it is returning `String` here.

barrybecker4: formatForIdx is returning a set of formatting functions for each column. Each of those functions takes as input the type of the data in the column and returns a string.  In the case of a timestamp column, a timestamp is the parameter and it returns a formatted datetime string.

HyukjinKwon: Doh, this is about writing. I thought it is about reading sorry.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -158,9 +158,13 @@ package object csv {
...
+          else dateFormatter.format(new Date(timestamp.asInstanceOf[Timestamp].getTime))
```
falaki: I was referring to putting braces for if else:
```scala
 if (timestamp == null) {
  nullValue
} else { 
  dateFormatter.format(new Date(timestamp.asInstanceOf[Timestamp].getTime)) 
}
```

barrybecker4:
I added the braces for the if/else. There was no style warning, so I didn't think it was required.

src/main/scala/com/databricks/spark/csv/package.scala
```diff
@@ -158,9 +158,17 @@ package object csv {
       // does not have to happen in the inner loop.
       val schema = dataFrame.schema
       val formatForIdx = schema.fieldNames.map(fname => schema(fname).dataType match {
-        case TimestampType => (timestamp: Any) =>
-          dateFormatter.format(new Date(timestamp.asInstanceOf[Timestamp].getTime))
-        case DateType => (date: Any) => dateFormatter.format(date)
+        case TimestampType => (timestamp: Any) => {
+          if (timestamp == null) {
+            nullValue
+          }
```
falaki: Sorry one more nit: 
```
} else {
  dateFormatter.format(new Date(timestamp.asInstanceOf[Timestamp].getTime))
}
```

barrybecker4: Isn't that what is there now? Maybe I am looking in the wrong place.

falaki: There is a new line after closing braces. We want them in the same line.

barrybecker4: ok. done.

PR #394, opened 2016-11-05
Fixed inferring of types under existence of malformed lines

digigek:
The type inference mechanism takes into account malformed lines, even though the user may have selected to drop them. In my opinion, this leads to faulty results, since the inferred types get distorted by unruly lines that may have more or less columns. Since those lines get dropped anyway during the creation of the dataframe, they shouldn't be taken into account during the type inference as well.


falaki:
Thanks for the fix. Would you please add unit tests?

digigek:
I will try to find some time to do it in the coming weeks.

digigek:
Hi, I have added 3 unit tests. When do you think, you could approve the merge?

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -53,14 +55,20 @@ private[csv] object InferSchema {
     StructType(structFields)
   }
 
-  private def inferRowType(nullValue: String, dateFormatter: SimpleDateFormat)
+  private def inferRowType(nullValue: String,
+                           header: Array[String],
+                           dropMalformed: Boolean = false,
```
falaki: Move this flag to the end of arg list.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -53,14 +55,20 @@ private[csv] object InferSchema {
...
+                           dateFormatter: SimpleDateFormat)
   (rowSoFar: Array[DataType], next: Array[String]): Array[DataType] = {
     var i = 0
-    while (i < math.min(rowSoFar.length, next.length)) {  // May have columns on right missing.
-      rowSoFar(i) = inferField(rowSoFar(i), next(i), nullValue, dateFormatter)
-      i+=1
+    if(header.length != next.length && dropMalformed) rowSoFar
```
falaki: style nit:
```
if () {
...
} else {
...
}
```

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -234,6 +232,8 @@ case class CsvRelation protected[spark] (
           escape = escapeVal,
           commentMarker = commentChar).parseLine(firstLine)
       } else {
+        val escapeVal = if (escape == null) '\\' else escape.charValue()
```
falaki: Why are these two lines part of this PR?

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -20,14 +20,12 @@ import java.text.SimpleDateFormat
 
 import scala.collection.JavaConversions._
 import scala.util.control.NonFatal
-
```
falaki: Do not remove these blank lines.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -18,8 +18,9 @@ package com.databricks.spark.csv.util
 import java.sql.Timestamp
 import java.text.SimpleDateFormat
 
-import scala.util.control.Exception._
+import com.databricks.spark.csv.readers.LineCsvReader
```
falaki: style: import order.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -53,14 +55,20 @@ private[csv] object InferSchema {
     StructType(structFields)
   }
 
-  private def inferRowType(nullValue: String, dateFormatter: SimpleDateFormat)
+  private def inferRowType(nullValue: String,
```
falaki: How does this function work? As far as I can see it is passed the flag for `dropMalformed` which is constant for all the rows (either `true` or `false`). Shouldn't the flag be set based on if the row is malformed or not?

falaki (commented):
Left some comments. Might be good to explain the logic in one line of comments.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
digigek: Fixed in next commit.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
digigek: Ooops sorry, this might have gotten out of my attention. Removed in next commit.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
digigek: Styled in next commit.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
digigek: Moved the flag to the end.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
digigek: The idea here is to know whether the user has used the DROPMALFORMED parse mode. Only in this case, the type inference should avoid taking into account the malformed lines. Whether a line is malformed or not is not the only criteria for doing the extra 'if' there. We need to check whether the user has Chosen the DROPMALFORMED Option as well. In this case, the parse mode is consistent both in the type inference and in the generation of the data Frame out of the CSV. 

This exact logic is depicted in my three unit tests as well. 

I also added a comment in the Code for avioding disambiguations.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
digigek: Removed the unused Import.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
falaki: Thanks for inline comments.

src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -248,7 +248,8 @@ case class CsvRelation protected[spark] (
       }
       if (this.inferCsvSchema) {
         val simpleDateFormatter = dateFormatter
-        InferSchema(tokenRdd(header), header, nullValue, simpleDateFormatter)
+        InferSchema(tokenRdd(header), header,
```
falaki: Style: Please do not break arguments on the second one. Either:
```
function(
    a, b, c)
```
Or
```
function(a,
    b,
    c)
```

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -53,14 +54,22 @@ private[csv] object InferSchema {
...
+                           header: Array[String],
```
falaki: Style: indent should be four spaces.

src/main/scala/com/databricks/spark/csv/util/InferSchema.scala
```diff
@@ -53,14 +54,22 @@ private[csv] object InferSchema {
...
+                           dateFormatter: SimpleDateFormat,
+                           dropMalformed: Boolean = false)
   (rowSoFar: Array[DataType], next: Array[String]): Array[DataType] = {
     var i = 0
-    while (i < math.min(rowSoFar.length, next.length)) {  // May have columns on right missing.
-      rowSoFar(i) = inferField(rowSoFar(i), next(i), nullValue, dateFormatter)
-      i+=1
+    if(header.length != next.length && dropMalformed) {
```
falaki: Style:
`if (header.length != next.length && dropMalformed) {`

src/test/resources/ages-alternative-malformed.csv
```diff
@@ -0,0 +1,6 @@
+
```
falaki: Please include one malformed line after the header as well. To have better coverage, please consider other types of malformed lines (like missing only one column or having extra?)

falaki (commented):
Thanks for addressing first round comments. I left a few more.

falaki:
Thanks. It is merged.

digigek:
Thanks for the review and the merge. How could I get the fix also into Spark2.x? Is there an automatic procedure with which critical patches from here are merged in Spark2.x?

Edit: After reviewing the code of the CSV datasource in Spark2.0, I got the answer myself. The code has been refactored heavily, so I will have to fix the bug individually there as well.

falaki:
I will publish a new minor release within next month. There might be other patches as well.

PR #395, opened 2016-11-07
delete ";" to match style to others

hayashidac:
I think it is better to delete ";" to match another scala source styles.


falaki:
Thanks for your pull request.

falaki:
Thanks for your patch.

PR #400, opened 2016-11-15
[SPARK-16512] Adding a insertNullOnErrors option.

rachelwarren:
This is the change that allows an option to render errors when parsing such as number format exceptions as nulls. 
It was in this pull request, https://github.com/databricks/spark-csv/pull/298 but I thought it would be cleaner to create a new one.


falaki:
@rachelwarren as you know this package is in maintenance mode and does not accept patches that modify the API. I suggest you send this patch to Apache Spark, where CSV data source has been in-lined since 2.x.

PR #417, opened 2017-03-16
Incorrectly checking ignoreLeadingWhiteSpace twice

omervk:


src/main/scala/com/databricks/spark/csv/CsvRelation.scala
```diff
@@ -63,7 +63,7 @@ case class CsvRelation protected[spark] (
     logger.warn(s"$parseMode is not a valid parse mode. Using ${ParseModes.DEFAULT}.")
   }
 
-  if ((ignoreLeadingWhiteSpace || ignoreLeadingWhiteSpace) && ParserLibs.isCommonsLib(parserLib)) {
+  if ((ignoreLeadingWhiteSpace || ignoreTrailingWhiteSpace) && ParserLibs.isCommonsLib(parserLib)) {
```
HyukjinKwon, line 66: Up to my knowledge, this option is not documented and does not work. I proposed the change in https://github.com/databricks/spark-csv/pull/306 but it was ported into Spark 2.x. So, I decided to close.

PR #420, opened 2017-03-28
Should read null properly when schema is lager than parsed tokens in PrunedScan

HyukjinKwon:
Currently running the codes below:

```scala
val schema = StructType(
  StructField("bool", BooleanType, true) ::
  StructField("nullcol", IntegerType, true) ::
  StructField("nullcol1", IntegerType, true) :: Nil)

new CsvParser()
  .withSchema(schema)
  .withUseHeader(true)
  .withParserLib(parserLib)
  .withParseMode(ParseModes.PERMISSIVE_MODE)
  .csvFile(sqlContext, boolFile)
  .select("bool", "nullcol")
  .collect()
```

with the data below:

```csv
bool
"True"
"False"

"true"
```

throws an exception as below:

```
java.lang.NumberFormatException: null
	at java.lang.Integer.parseInt(Integer.java:542)
	at java.lang.Integer.parseInt(Integer.java:615)
	at scala.collection.immutable.StringLike$class.toInt(StringLike.scala:272)
	at scala.collection.immutable.StringOps.toInt(StringOps.scala:30)
	at com.databricks.spark.csv.util.TypeCast$.castTo(TypeCast.scala:57)
	at com.databricks.spark.csv.CsvRelation$$anonfun$buildScan$6.apply(CsvRelation.scala:196)
	at com.databricks.spark.csv.CsvRelation$$anonfun$buildScan$6.apply(CsvRelation.scala:174)
```

This is related with https://github.com/apache/spark/pull/15767. Please refer the description there.

### DIFF
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -57,7 +57,7 @@ object TypeCast {
 
     // If the given column is not nullable, we simply fall back to normal string
     // rather than returning null for backwards compatibility. Note that this case is
-    // different with Spark's internal CSV datasource.
+    // different with Spark's internal CSV datasource which throws an exception in this case.
     val isNullValueMatched = datum == nullValue && nullable
 
     // If `treatEmptyValuesAsNulls` is enabled, treat empty strings as nulls.
@@ -67,7 +67,7 @@ object TypeCast {
     // In this case, they are treated as nulls.
     val isNullDatum = datum == null
 
-    if (isNullValueMatched || shouldTreatEmptyValuesAsNulls || isNullDatum){
+    if (isNullValueMatched || shouldTreatEmptyValuesAsNulls || isNullDatum) {
       if (!nullable) {
         throw new RuntimeException(s"null value found but field $name is not nullable.")
       }
--- a/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
@@ -34,21 +34,43 @@ object TypeCast {
    * Currently we do not support complex types (ArrayType, MapType, StructType).
    *
    * For string types, this is simply the datum. For other types.
-   * For other nullable types, this is null if the string datum is empty.
+   * For other nullable types, returns null if it is null or equals to the value specified
+   * in `nullValue` option. If `treatEmptyValuesAsNulls` is set, it also returns null for
+   * empty strings.
    *
    * @param datum string value
-   * @param castType SparkSQL type
+   * @param name field name in schema.
+   * @param castType data type to cast `datum` into.
+   * @param nullable nullability for the field.
+   * @param treatEmptyValuesAsNulls a flag to indicate if empty strings should be treated as null.
+   * @param nullValue the string value that represents null.
+   * @param dateFormatter date formatter that uses to parse dates and timestamps.
    */
   private[csv] def castTo(
       datum: String,
+      name: String,
       castType: DataType,
       nullable: Boolean = true,
       treatEmptyValuesAsNulls: Boolean = false,
       nullValue: String = "",
       dateFormatter: SimpleDateFormat = null): Any = {
-    if (datum == nullValue &&
-      nullable ||
-      (treatEmptyValuesAsNulls && datum == "")){
+
+    // If the given column is not nullable, we simply fall back to normal string
+    // rather than returning null for backwards compatibility. Note that this case is
+    // different with Spark's internal CSV datasource.
+    val isNullValueMatched = datum == nullValue && nullable
+
+    // If `treatEmptyValuesAsNulls` is enabled, treat empty strings as nulls.
+    val shouldTreatEmptyValuesAsNulls = treatEmptyValuesAsNulls && datum == ""
+
+    // `datum` can be null when some tokens were inserted when permissive modes via `PrunedScan`.
+    // In this case, they are treated as nulls.
+    val isNullDatum = datum == null
+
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -34,21 +34,43 @@ object TypeCast {
...
+    // rather than returning null for backwards compatibility. Note that this case is
+    // different with Spark's internal CSV datasource.
+    val isNullValueMatched = datum == nullValue && nullable
+
+    // If `treatEmptyValuesAsNulls` is enabled, treat empty strings as nulls.
+    val shouldTreatEmptyValuesAsNulls = treatEmptyValuesAsNulls && datum == ""
+
+    // `datum` can be null when some tokens were inserted when permissive modes via `PrunedScan`.
+    // In this case, they are treated as nulls.
+    val isNullDatum = datum == null
```
HyukjinKwon: Here, the actual cases added are, 

- `datum == null && !nullable` : returns `null`. (it was NEP before).
- `datum == null && !nullable` : throws an exception. (it was NEP before).
- `treatEmptyValuesAsNulls && datum == && !nullable` : throws an exception. (it was trying to set `null` to non-nullable field.

These should be non-behaviour changes.

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
```diff
@@ -61,62 +61,72 @@ class TypeCastSuite extends FunSuite {
...
+
+    val exception2 = intercept[RuntimeException]{
+      TypeCast.castTo("", "_c", StringType, nullable = false, treatEmptyValuesAsNulls = true)
+    }
+    assert(exception2.getMessage.contains("null value found but field _c is not nullable"))
+
+    val exception3 = intercept[RuntimeException]{
+      TypeCast.castTo(null, "_c", StringType, nullable = false)
+    }
+    assert(exception3.getMessage.contains("null value found but field _c is not nullable"))
```
HyukjinKwon: Here is the actual test cases added.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
```diff
@@ -34,21 +34,43 @@ object TypeCast {
...
       nullValue: String = "",
       dateFormatter: SimpleDateFormat = null): Any = {
-    if (datum == nullValue &&
-      nullable ||
-      (treatEmptyValuesAsNulls && datum == "")){
+
+    // If the given column is not nullable, we simply fall back to normal string
+    // rather than returning null for backwards compatibility. Note that this case is
+    // different with Spark's internal CSV datasource.
+    val isNullValueMatched = datum == nullValue && nullable
```
HyukjinKwon: I think this case is different with Spark currently but did not change the behaviour here.
In case of Spark, it throws an exception in this case rather than failing back to string as below:

```scala
val schema = StructType(
  StructField("a", IntegerType, true) ::
  StructField("b", IntegerType, false) :: Nil)

spark.read.schema(schema)
  .option("header", "false")
  .option("nullValue", "2")
  .option("mode", "FAILFAST")
  .csv(Seq("1,2").toDS())
  .show()
```

prints

```
java.lang.RuntimeException: null value found but field b is not nullable.
	at org.apache.spark.sql.execution.datasources.csv.UnivocityParser.org$apache$spark$sql$execution$datasources$csv$UnivocityParser$$nullSafeDatum(UnivocityParser.scala:179)
	at org.apache.spark.sql.execution.datasources.csv.UnivocityParser$$anonfun$makeConverter$3.apply(UnivocityParser.scala:104)
	at org.apache.spark.sql.execution.datasources.csv.UnivocityParser$$anonfun$makeConverter$3.apply(UnivocityParser.scala:103)
	at org.apache.spark.sql.execution.datasources.csv.UnivocityParser.org$apache$spark$sql$execution$datasources$csv$UnivocityParser$$convert(UnivocityParser.scala:219)
	at org.apache.spark.sql.execution.datasources.csv.UnivocityParser.parse(UnivocityParser.scala:191)
```

In case of this library,

```scala
val schema = StructType(
  StructField("a", IntegerType, true) ::
  StructField("b", IntegerType, false) :: Nil)

new CsvParser()
  .withSchema(schema)
  .withUseHeader(false)
  .withParseMode(ParseModes.FAIL_FAST_MODE)
  .withNullValue("2")
  .csvRdd(sqlContext, sqlContext.sparkContext.parallelize(Seq("1,2")))
  .show()
```

prints below:

```
+---+---+
|  a|  b|
+---+---+
|  1|  2|
+---+---+
```

HyukjinKwon: I intended to not break this behaviour in this library but I am willing to match this to Spark's internal CSV one if you think so.

src/test/scala/com/databricks/spark/csv/util/TypeCastSuite.scala
HyukjinKwon: Here are the actual test cases added.

src/main/scala/com/databricks/spark/csv/util/TypeCast.scala
HyukjinKwon, line 68: Here, the actual cases added are, 

- `datum == null && !nullable` : returns `null`. (it was NEP or `java.lang.NumberFormatException: null` before).
- `datum == null && !nullable` : throws an exception. (it was NEP or `java.lang.NumberFormatException: null` before).
- `treatEmptyValuesAsNulls && datum == && !nullable` : throws an exception. (it was trying to set `null` to non-nullable field.

These should be non-behaviour changes.

HyukjinKwon:
(Two comments were folded by the last commit but still valid to explain the changes here)

PR #472, opened 2018-12-13
Write only one header option (headerOnFirstFile)

hangingman:
I append `headerOnFirstFile` option. 
If this option enabled, it will generate 1-header only. ( according to #317 )
If this option disabled, spark-csv will generate header per file. Defaut is false.


falaki:
@hangingman this project is in maintenance mode after the code was contributed to Apache Spark. Please submit your contribution to Apache Spark.

